PR #8, opened 2022-10-12
Add Dart CLI template

leograba:
This PR goal is to add Dart support, using a CLI template created with the `dart create -t console` command.

It has some limitations:

* Directory name: in the templates directory I'm using `dart_cli` with all lowercase and underscores due to https://dart.dev/tools/pub/pubspec#name. If that is an issue I can update it, but the directory name will not be compliant with the Dart pubspec.
    * When a project is created from template, the extension does not allow underscores, so the project created from template will not comply with the pubspec.
* To start cross-debugging, the user will be prompted for a "VM URI". This initial version does not automate this process because the URI has a unique token that is randomly created every time `dart` starts the application. The URI is printed to a focused terminal, so the user only has to copy/paste.
* The cross debugger can pause and step, but it does not find the source code, so breakpoints and possibly other things are not working.
* The Dart extension has support for hot reload, but it is not implemented in this initial cross debug integration.
* The Dart logo added with this PR is taken as-is from https://dart.dev/brand. It doesn't seem to be very similar to the existing ones. I'm not very good at editing logos, though.


microhobby:
@andreriesco would you like to continue this as a community maintainer of this template? Let me know

microhobby:
@andreriesco could you please help with this review?

microhobby:
@leograba could you please change the request to `main` branch to `dev` branch. I will remove the `main` branch since we will maintain the `bookworm` and `bullseye` branches now. So, we expect to have all the new commits being sent to the `dev` branch. Thanks in advance.

leograba:
I think I'll drop this PR, I got a sense of how to write an extension and I don't remember much anymore so it would take time to resume work here.

PR #14, opened 2022-11-03
scripts: Add SuppressMessageAttribute

microhobby:
We are using some methods that are warned as not recommended.
But they are needed by our use case. On future we can check for 
better solutions, although the warnings are more focused on the script being supported on cross platform and we are only using it on Linux.


microhobby:
@andreriesco could you review and test this? Thanks

andreriesco:
LGTM

PR #23, opened 2022-12-12
tests: fix dotnetUno successDebugMessage

andreriesco:
Related-to: TIE-648

### DIFF
--- a/templates.json
+++ b/templates.json
@@ -13,8 +13,9 @@
                 "debuggers": [
                     ".NET Core Launch (console)", 
                     ".NET Core Attach", 
-                    ".NET Attach Torizon ARMv8", 
-                    "Torizon x86", "Torizon ARMv7", 
+                    ".NET Attach Torizon arm64",
+                    "Torizon x86",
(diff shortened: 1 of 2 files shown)


templates.json
```diff
@@ -63,7 +63,7 @@
                 "debuggerConsole": null,
                 "mainFile": "Program.cs",
                 "mainFileBreakpointLine": 29,
-                "successDebugMessage": "Loaded 'Anonymously Hosted DynamicMethods Assembly'."
+                "successDebugMessage": "System.Linq.Expressions.dll'. Skipped loading symbols. Module is optimized and the debugger option 'Just My Code' is enabled."
```
microhobby, line 66: Maybe it's better to add some `Console.Writeline("Hello world")` to the template to be sure that this will not be changed by external libs ...

microhobby (changes requested):

templates.json
andreriesco: Ok, I will also add a "Hello World" on the cppQML template for the same purpose as this one.

PR #26, opened 2022-12-19
settings: tasks: launch: Add the torizon_run_as

microhobby:
This is needed if exist the need to run the application with another user, root for example.

Signed-off-by: Matheus Castello <matheus.castello@toradex.com>
Related-to: TIE-729


microhobby:
@andreriesco FYI

PR #28, opened 2023-01-03
Add cspell check

microhobby:
To improve the quality of the software we are adding a spell checker to the script files. At the moment the checker only checks comments and strings (some lines can be ignored, for example if they are arguments for input to a program)


microhobby:
@andreriesco FYI

PR #36, opened 2023-01-17
assets/tasks/common.json: Set GPU variable on debug

lucas-akira:
This is necessary for SoMs that use a Vivante GPU, which uses weston-vivante instead of weston.

Signed-off-by: morishitalucas@gmail.com


microhobby:
LGTM

microhobby:
@lucas-akira thanks for your contribution!

PR #44, opened 2023-03-23
scripts: torizonPackages: Change TORIZON_ARCH armhf check

lucas-akira:
Change TORIZON_ARCH comparison from 'arm' to 'armv7l'.

### DIFF
--- a/scripts/torizonPackages.ps1
+++ b/scripts/torizonPackages.ps1
@@ -22,7 +22,7 @@ if ($TORIZON_ARCH -eq "aarch64") {
     $TORIZON_ARCH = "arm64" 
 }
 
-if ($TORIZON_ARCH -eq "arm") { 
+if ($TORIZON_ARCH -eq "armv7l") { 
     $TORIZON_ARCH = "armhf" 
 }
 


scripts/torizonPackages.ps1
```diff
@@ -22,7 +22,7 @@ if ($TORIZON_ARCH -eq "aarch64") {
     $TORIZON_ARCH = "arm64" 
 }
 
-if ($TORIZON_ARCH -eq "arm") { 
+if ($TORIZON_ARCH -eq "armv7l") {
```
microhobby: which hardware you are using to test it? This one is trick, it's only `arm` because this should cover the `armv7l` and the `armhf`cases...

But let me know, if this is not working in a generic way, we should be more specific...

lucas-akira: I should have made this clearer, my fault.

I'm using an Apalis i.MX6Q, and in this case, in the original code, before the `if` statement in question:

- The value of `TORIZON_ARCH` is `armv7l`
- The `if` statement is false i.e. string `armv7l` is not equal to `arm`
- `TORIZON_ARCH` doesn't change to `armhf`, so the script tries to install packages with suffix `:armv7l`, which doesn't exist in the Debian feeds.

For instance, trying to install `libserial-dev` from `torizonPackages.json` (On a C/C++ ApolloX project):
```
{
    "deps": [
    ],
    "devDeps": [
        "libserial-dev"
    ]
}

```

Results in:
```
Executing task: DOCKER_HOST="" docker build -f /home/user/toradex/apollox_projects/cppApolloxTest3/Dockerfile.sdk /home/user/toradex/apollox_projects/cppApolloxTest3 -t cross-toolchain-arm-cpp_apollox_test3 --build-arg IMAGE_ARCH=arm  
[...]
> [3/4] RUN apt-get -q -y update &&     apt-get -q -y install  libserial-dev:armv7l     &&     apt-get clean && apt-get autoremove &&     rm -rf /var/lib/apt/lists/*:    
#0 0.541 Get:1 http://deb.debian.org/debian bullseye InRelease [116 kB]  
[...]
#0 5.651 E: Unable to locate package libserial-dev:armv7l
```

microhobby: Hmm yeah, I see. LGTM

PR #46, opened 2023-04-17
Fix on python templates

andreriesco:


python3Pyside2QML/docker-compose.yml
```diff
@@ -56,7 +56,7 @@ services:
     ]
 
   weston:
-    image: torizon/weston${GPU}:3
+    image: torizon/weston${GPU}:2
```
microhobby, line 59: Use the 3 cause some issue in the Torizon 6 + pyside?
Because this is only set in the Bookworm branch, in the bullseye it's continue 2

andreriesco: Yes, and just changing to the 2 fixed the problem, on TC 6.

microhobby:
This was cherry-picked since there was not synced with dev latest commits.

PR #52, opened 2023-04-21
Add new C++ and Rust templates for GUI applications written in Slint

tronical:


tronical:
I've tried this on a imx8 on a dahlia board and it works for me. The docker file is a little complex, and it also takes a fair while to build. The upside though is that since the slint building part is the same for the SDK, the deploy image for the SDK, as well as the production, once it's built those layers are reused.

tronical:
Regarding the logo: Do you know by chance if SVGs are supported, too, by your extension? That would look better than the PNG file for the Slint logo.

Edit: I changed it to use an SVG - works fine and looks much better :)

tronical:
I've taken the liberty of also adding the Rust template into the same PR.

Happy to make any kind of squashes/rebases/etc. as requested. I can also split it again into tow PRs if you'd like.

microhobby:
Hey @tronical thanks for the contributions! I will take a look soon.

microhobby:
@tronical sorry I only see this now, but all the new development should be merged into the `dev` branch. Could you please use the `dev` branch as base for your development?

Tip: I think that will be easier to you get your commits and cherry-pick to a fresh rebase from dev than try to rebase `dev` into `bookworm`

tronical:
Sure, I'll rebase onto dev.

tronical:
Done. There were no conflicts. Did dev diverge from bookworm in a way that would require making changes to the docker files, etc?

microhobby:
@tronical oh cool, not differ in content, only in commit hashes, so I thought that this will be an issue. But good to know. Thanks!

cppSlint/.conf/deps.json
```diff
@@ -0,0 +1,12 @@
+{
+    "packages": [
+        "openssh-client",
+        "sshpass",
+        "build-essential",
+        "make",
+        "gdb",
+        "gdb-multiarch",
+        "g++",
+        "netcat-traditional"
```
microhobby: This package is only needed from Qt because we do some port redirect to use the Qt Creator as debugger. Are you using it for something in the Slint template?

cppSlint/.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,17 @@
+{
+    "configurations": [
+        {
+            "name": "Linux",
+            "includePath": [
+                "${workspaceFolder}/**",
+                "/usr/include/**"
```
microhobby, line 7: There is some way to download and configure here the Slint headers, like the `appwindow.h` for example?

microhobby: The cpp linter works locally, and we should have it locally to set in the `includePath`.
Maybe to add a task entry to be executed in the project open to download the headers...

cppSlint/Dockerfile.sdk
```diff
@@ -0,0 +1,106 @@
...
+
+RUN curl https://sh.rustup.rs -sSf | sh -s -- -y
+RUN rustup target add $(cat /rust-toolchain-arch.txt)
+
+# Install Slint build dependencies (libxcb, etc.)
+RUN \
+    --mount=type=cache,target=/var/cache/apt \
+    CROSS_TOOLCHAIN_ARCH=$(cat /cross-toolchain-arch.txt) && \
+    apt-get update && \
+    apt-get install --assume-yes libfontconfig1-dev:$CROSS_TOOLCHAIN_ARCH libxcb1-dev:$CROSS_TOOLCHAIN_ARCH libxcb-render0-dev:$CROSS_TOOLCHAIN_ARCH libxcb-shape0-dev:$CROSS_TOOLCHAIN_ARCH libxcb-xfixes0-dev:$CROSS_TOOLCHAIN_ARCH libxkbcommon-dev:$CROSS_TOOLCHAIN_ARCH python3 clang libstdc++-10-dev:$CROSS_TOOLCHAIN_ARCH && \
```
microhobby: `libstdc++-10-dev` has no candidate for Debian `bookworm`. Is possible to use the `libstdc++-11-dev`? https://packages.debian.org/search?suite=bookworm&section=all&arch=any&searchon=names&keywords=libstdc%2B%2B-

cppSlint/Dockerfile.sdk
```diff
@@ -0,0 +1,106 @@
...
+    rm -rf /var/lib/apt/lists/*
+
+# Default to the Skia backend for best performance
+ENV SLINT_BACKEND=winit-skia
+# Default to Slint running in fullscreen
+ENV SLINT_FULLSCREEN=1
+
+USER torizon
+
+WORKDIR /app
```
microhobby, line 106: yeah, this take some time to build. Maybe should be good to you guys provide some pre-built Docker images to use as base, like `slint/slint-ui` for `arm64v8` and `arm32v7`. This should greatly improve build speed.

assets/img/slint-logo-square-light.svg
microhobby, line 1: For maintaining the consistency, I would like to propose to use the follow images:
slintrust
slintcpp

Let me know if this works for you.

microhobby (changes requested):

microhobby:
@tronical for now I only checked the C++ one, next week we will check the Rust template.

cppSlint/Dockerfile.sdk
microhobby: The comment above is only a suggestion, this will not block the merge.

cppSlint/.vscode/settings.json
```diff
@@ -0,0 +1,19 @@
...
+  "torizon_debug_port2": "2232",
+  "torizon_debug_port3": "",
+  "terminal.integrated.defaultProfile.linux": "bash",
+  "torizon_gpu": "",
+  "torizon_arch": "",
+  "wait_sync": "1",
+  "torizon_run_as": "torizon",
+  "torizon_app_root": "/home/torizon",
+  "tcb.packageName": "__change__",
+  "tcb.version": "early-access"
```
microhobby, line 18: Did you do not want to add here the:

```json
  "slint.preview.providedByEditor": true
```

I was testing here and this integrates very well instead of the default preview that uses a fullscreen window

cppSlint/.conf/deps.json
tronical: Indeed, that's a leftover from copy & paste. Will remove - Thanks :)

cppSlint/.vscode/c_cpp_properties.json
tronical: Hm, `appwindow.h` is a generated header file (from `appwindow.slint`), so I'm not sure where that could be downloaded from/to. The Slint C++ API headers are inside the docker image and installed into /usr/include.

How do you handle this with other C/C++ libraries?

cppSlint/.vscode/settings.json
tronical: I'd prefer to stay with the defaults here. The side-by-side works, but in terms of screen real estate that remains left for the code editor, I prefer the separate preview window.

cppSlint/Dockerfile.sdk
tronical: Yes, that's the correct choice, I'll fix that. When I started, the 3-bookwork tag only had libstdc++10-dev, but it seems that meanwhile you moved the tag to a newer Debian snapshot where 10 was replaced with 11.

cppSlint/Dockerfile.sdk
tronical: Yeah, it's tempting to do that. But it's also a bit of work to maintain them in parallel your images. When you update your 3-bookworm tag, we need to update our images, too. Another option is to entirely deviate from Torizon, but then we might run into issues where our container creates run-time library dependencies that aren't present in the torizon deployment containers.

assets/img/slint-logo-square-light.svg
tronical: Excellent idea! Let me see if I can integrate those tag images into the .svg, otherwise we can use your png files :)

tronical: Alright, I can't see a trivial way - so I pushed a commit with your png versions. Thanks again :)

cppSlint/.vscode/settings.json
microhobby: Is there a way to not show the window in full screen, then? My concern is that it makes users, that uses single screen, confused.

rustSlint/src/main.rs
```diff
@@ -0,0 +1,15 @@
...
+fn main() -> Result<(), slint::PlatformError> {
+    let ui = AppWindow::new()?;
+
+    let ui_handle = ui.as_weak();
+    ui.on_request_increase_value(move || {
+        let ui = ui_handle.unwrap();
+        ui.set_counter(ui.get_counter() + 1);
+    });
+
+    println!("Hello, Toradex!");
```
microhobby: I think here should be `Hello Torizon!`

microhobby (changes requested):

cppSlint/.vscode/settings.json
microhobby: ok, I tried again and it's showing a normal window, resolving also this one

cppSlint/.vscode/c_cpp_properties.json
microhobby: We usually instruct users to install the reference dev package `yourlib-dev`, or download the headers to have a local reference. As I said, the linter runs locally, I mean outside the SDK container.

But from my tests the `build-debug-amd64-local`, already get the `build/x86_64/_deps/slint-src/api/cpp/include` folder that has the API headers. So, LGTM

microhobby:
@tronical thanks a lot for the changes, I check also the rust one and apart from the small detail here https://github.com/toradex/vscode-torizon-templates/pull/52#discussion_r1188081030, LGTM. As soon as you change this detail we will merge it, thank you.

rustSlint/src/main.rs
tronical: Ah yes, well spotted!

tronical:
Thanks for the review! I pushed the fix.

PR #56, opened 2023-04-26
share WSL ports Series

microhobby:
These are some changes and fixes for the share WSL ports scenario.
In some cases the `run-docker-registry` can not work telling that the `5002` is already in use, but the process that is using it is the Windows firewall utils. So this series adds the configurations and fixes to run the registry as soon the delete the old port from firewall.


andreriesco:
LGTM

Issue #51, opened 2023-04-20
Limited Support for graphical acceleration on downstream-based modules


brunoaamello:
There are templates that use the container [wayland-base](https://hub.docker.com/r/torizon/wayland-base) as a base for their application containers, regardless of SoC/SoM. 
This results in limited hardware acceleration support for downstream-based modules (i.MX 8/i.MX 8M/i.MX 8X).

The downstream-based modules should use the container [wayland-base-vivante](https://hub.docker.com/r/torizon/wayland-base-vivante) instead, to allow hardware acceleration within the container.
As it is right now, the templates have hardware accelerated compositing, in the separate [weston-vivante](https://hub.docker.com/r/torizon/weston-vivante) container, but the OpenGL provider in the application container itself is [LLVMpipe](https://docs.mesa3d.org/drivers/llvmpipe.html), a software implementation.

The affected templates are:
- [C++ Qt 6 QML Application](https://github.com/toradex/vscode-torizon-templates/tree/bookworm/cppQML)
- [Node.js JavaScript Electron Application](https://github.com/toradex/vscode-torizon-templates/tree/bookworm/nodeElectron)
- [Python 3 Pyside 2(Qt5) QML](https://github.com/toradex/vscode-torizon-templates/tree/bookworm/python3Pyside2QML)

The fix can be as simple as using a build arg for the GPU or may require further changes to these templates.

microhobby:
This is blocked because we are having some issues with the `torizon/wayland-base-vivante:3` packages when trying to install `qt6-base-dev`:

```
root@ef1e75ce2d6c:/# apt install qt6-base-dev
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 qt6-base-dev : Depends: libgl-dev
E: Unable to correct problems, you have held broken packages.
```

From #57

microhobby:
For the Qt6 we need to wait until we have the packages build with OpenGL ES support.
This is under WIP by the Toradex Debian Package team.

microhobby:
@tronical Slint supports OpenGL ES? We need to toggle something to work?
I'm trying to run the Slint CPP with the Vivante GPU libraries and I'm getting:

```
[...]

PR #57, opened 2023-04-30
cppQML: Add the GPU vendor prefix to the dockerfiles

microhobby:
This will allow us to build the image using the correct base image for the specific GPU vendor.

Related-to: #51


andreriesco:
LGTM

PR #58, opened 2023-05-09
Update the community templates for Bookworm

andreriesco:


cppConsole/.vscode/tasks.json
```diff
@@ -123,6 +123,7 @@
             "args": [
                 "docker",
                 "build",
+                "--pull",
```
microhobby: hmm, why are we not using the commons on `cppConsole`?

cppQML/.vscode/tasks.json
```diff
@@ -375,6 +375,7 @@
             "args": [
                 "docker",
                 "build",
+                "--pull",
```
microhobby: `cppQML` also not using commons 🙃

rustConsole/.vscode/tasks.json
```diff
@@ -37,6 +37,7 @@
             "args": [
                 "docker",
                 "build",
+                "--pull",
```
microhobby: `rustConsole` also not using commons?

javaForms/docker-compose.yml
```diff
@@ -36,7 +36,7 @@ services:
     ]
 
   weston:
-    image: torizon/weston${GPU}:3
+    image: torizon/weston${GPU}:2
```
microhobby, line 39: using the bullseye on the app and bookworm on weston works?

python3Pyside2QML/src/main.py
```diff
@@ -8,16 +8,7 @@
...
-    if(os.path.exists(os.path.expanduser("~/app/mainwindow.qml"))):
-        # Path of the qml source code file inside the debug container on the board, to debug the application remotelly on the board
-        qurlPath = os.path.expanduser("~/app/mainwindow.qml")
-    elif (os.path.exists("/app/mainwindow.qml")):
-        # Path of the qml source code file inside the release container on the board, to run the release (production) application on the board
-        qurlPath = "/app/mainwindow.qml"
-    elif(os.path.exists("src/mainwindow.qml")):
-        # Path of the qml source code file on host, to debug the application locally
-        qurlPath = "src/mainwindow.qml"
+    qurlPath = qurlPath = "src/mainwindow.qml"
```
microhobby, line 11: this also works when running from the image generated from production Dockerfile?

rust-console/src/main.rs
```diff
@@ -1,3 +1,3 @@
 fn main() {
-    println!("Hello, Toradex!");
+    println!("Hello Torizon!");
```
microhobby: please rebase your branch on top of `dev` and also take a look on the slint templates, I think that we need to stadardize there too.

cppConsole/.vscode/tasks.json
andreriesco: Those are tasks for building the SDK container, which weren't on the commons, I don't know why

andreriesco: I will take a look if they can go to common

javaForms/docker-compose.yml
andreriesco: Yes, and for now it is necessary because there is a bug on the `weston-vivante:3` that makes this template and the .NET Avalonia GTK template to not work. But in the future, one this bug is fixed it can be moved back to 3.

python3Pyside2QML/src/main.py
andreriesco: Actually it had a little issue there. I fixed it in the latest rebase now.

cppConsole/.vscode/tasks.json
andreriesco: I added a commit with this now

rustConsole/.vscode/tasks.json
andreriesco: I added a commit with this now

cppQML/.vscode/tasks.json
andreriesco: I added a commit with this now

microhobby:
LGTM

PR #61, opened 2023-05-19
Add GPU build arg support & Fixes for GPU HW Acceleration

### LINKED ISSUE #51: Limited Support for graphical acceleration on downstream-based modules

microhobby:
For HW that does not use upstream libs, AKA proprietary GPU Vendor libs, the GPU build arg is usefull to tell the Dockerfile the correct Docker Image to use and which libs install.

Fixes: #51


cppSlint/Dockerfile
```diff
@@ -11,11 +11,17 @@ ARG IMAGE_ARCH=
 ##
 ARG APP_EXECUTABLE=__change__
 
+##
```
tronical, line 14: Would be awesome if this could also be added to the Slint Rust template. I can also do that after this PR (or make a patch for this branch).

microhobby: Sure, I will add here on my pipeline to include also the `rustSlint` on these changes.

microhobby:
Superseded by #67

PR #62, opened 2023-05-21
fix: rustConsole: issues in production builds

### LINKED ISSUE #53: console-rust template doesn't build with production Dockerfile
The copy instruction to deploy the target binary in rust-console's Dockerfile makes two incorrect assumptions:

` COPY --from=Build /app/debug /app`

1. That the production build is a debug build.
2. That the build was a "native" build based on the default target platform.

However the earlier build instructions run `cargo build --target XXX` and consequently the binaries end up in `target/XXX/debug`.

I think the production build should do a release build, and some trickery is needed to copy it out of the `target/XXX/release/` directory.

(I need that trickery anyway in the Slint template, so if I find a good solution I'll post it here)

lucshl:
Fixes #53 by applying the given suggestions.

Thanks for the tips @tronical. BTW great work on the Slint templates :-)


microhobby:
@skkywalker could you please add your changes based on the `dev` branch? All new development is going to the `dev` branch. The `bookworm` branch is the default to be easier to the extension to clone the default branch. Sorry for the inconvenient.

lucshl:
Sorry, I'll close this one in favor of [!64](https://github.com/toradex/vscode-torizon-templates/pull/64).

PR #66, opened 2023-05-30
fix: rustConsole: fix output binary's name in launch.json

lucas-tdx:
After [!64](https://github.com/toradex/vscode-torizon-templates/pull/64), `Cargo.toml` was changed to name the output binary file to be the same as the project's name. This breaks the launch scripts, which expects the binary to be called `main`.

See https://github.com/toradex/vscode-torizon-templates/pull/64/files#diff-56d1d15f0241d43d317e04f6f40bf600bddb6e7d4e8e4a0ade43df4286f8a7bfL2 for the change that introduced the issue.


andreriesco:
LGTM

microhobby:
LGTM

PR #68, opened 2023-06-13
tasks: Fix password with special characters

microhobby:
If we have a password with special characters like $, the task will fail because the password is not being passed correctly to the command line. This patch fixes this issue by adding single quotes around the password.


microhobby:
LGTMS

PR #71, opened 2023-06-14
scripts: initWorkspace: Enforce right permissions to .conf/id_rsa

microhobby:
In some scenarios we can get an debug id_rsa file with the wrong permissions, so let's enforce the 0400 like we do on vs code side.


andreriesco:
LGTM

PR #72, opened 2023-06-26
first attempt of next.js template

escherstair:
this is the first attempt to create a template for Next.js project.
[What is Next.js?](https://nextjs.org/docs#what-is-nextjs)

Since Next.js is based over Node.js, I started from Node.js template.
At this stage this template doesn't work yet, but I opened this PR so that the community can review it and speed up the process.

The Next.js project creates a server (with UI too), which can be accessed on [http://localhost:3000](http://localhost:3000) with the browser to see the result.
For this reason I don't know if Next.js container must depend on weston-vivante, or it's only Chromium/Cog that must have this dependency.

Feel free starting the review of this PR.


microhobby:
@escherstair I'm changing the base of the MR for the `dev` branch, ok? The `bookworm` is the default production branch, and we should add new stuff on the `dev` branch first. I will continue the review basing from the `dev` branch.

escherstair:
@microhobby I did some improvements to the files.
I'm almost sure that something doesn't work as expected, but I did my best.

For the debug experience, I took [this official link](https://nextjs.org/docs/pages/building-your-application/configuring/debugging) as an inspiration. But I'm not 100% convinced of the result.

[This](https://github.com/vercel/next.js/tree/canary/examples/with-docker) should be the configuration to be used with docker, but I'm not sure if all of the steps can be integrated in torizon workflow.

microhobby:
@escherstair I take a look on how the NextJS works and I was able to attach to the server side using the changes from: a93135811636c08d24a9317d40e289cc1ecfde7c

Like running the debug target and then adding a breakpoint to the `hello.ts` return for the api.

nextJSTypeScript/Dockerfile
```diff
@@ -0,0 +1,139 @@
...
+# Board GPU vendor prefix
+##
+ARG GPU=
+# ARGUMENTS --------------------------------------------------------------------
+
+
+
+# BUILD ------------------------------------------------------------------------
+FROM --platform=linux/${IMAGE_ARCH} \
+    torizon/wayland-base${GPU}:${BASE_VERSION} AS Build
```
microhobby, line 29: There is some specific reason why are you using the `torizon/wayland-base` instead `torizon/debian`?

escherstair: I'm not sure about this.
Since Next.js has both a Server side and a client side, and the GUI is accessed through Chromium (typically), I thought that `torizon/wayland-base` would be necessary to support GPU-accelerated rendering.
But maybe Chromium container is enough to do this, and `torizon/debian` is the right choice.

escherstair:
> @escherstair I take a look on how the NextJS works and I was able to attach to the server side using the changes from: [a931358](https://github.com/toradex/vscode-torizon-templates/commit/a93135811636c08d24a9317d40e289cc1ecfde7c)
> 
> Like running the debug target and then adding a breakpoint to the `hello.ts` return for the api.

Hi @microhobby 
I pulled your changes and I tried to do the same. I use ApolloX in WSL (not a native Linux machine).
If I run the debug target, I get the following messages from the task `run-nextjs-local`
```
 *  Executing task: npm run dev 


> hello_world@1.0.0 dev
> next dev

- ready started server on 0.0.0.0:3000, url: http://localhost:3000
- event compiled client and server successfully in 185 ms (18 modules)
- wait compiling...
- event compiled client and server successfully in 130 ms (18 modules)
```
The messages seems ok to me, and the server runs properly because I can access using a browser to http://localhost:3000 on my Windows machine.
When I access the page, the following messages are added to `run-nextjs-local` terminal
```
- wait compiling /api/hello (client and server)...
- event compiled successfully in 157 ms (48 modules)
```
But the debugger is not able to attach:
the icon 
immagine
stays in waiting forever, and I never see the debug "buttons" 
immagine

I suspect that the debug port is somehow closed, or cannot be used.
Is this a WSL-only issue?

microhobby:
@escherstair I do not think so, I'm also using on WSL. This next.js project seems to have a lot of non expect behavior when you need to use the node parameters to debug, so I'm having a hard time to setup a remote debugging. For example 4 days ago they fix this: https://github.com/vercel/next.js/commit/70aad4a2b5b34fbbae006c477f96c54118da6381

escherstair:
I understand. I'll involve one of my colleagues in case he can help.
@microhobby Which versions of Node/Next did you install on WSL?
Let me know what you need me to do.

microhobby:
@escherstair I pushed some changes, I'm using the `patch-package` to patch some issues from NextJS when trying to use the node `--inspect`. Also, now we are trying to attach to all the 3 processes using a compound. So, when you try to use it, select the target `Next.js Local`:


https://github.com/toradex/vscode-torizon-templates/assets/2633321/5a0463eb-5d0b-4409-8c35-b35fe4b70e46

microhobby:
@escherstair I'm also adding a compound for the remote debugger:

https://github.com/toradex/vscode-torizon-templates/assets/2633321/f13317e6-788d-4af6-80c1-2e67484e5dc7

in the same way, this try to attach to the 3 process created by the NextJS. Take a look and check if this works for you use case.

Obs: I'm still not satisfied with the solution, for NextJS to run correctly it seems to need to run `npm install` on the remote target, so NPM gets the right libraries for armv8. But it takes a while each time I have to run the debug session. If you have ideas to improve this, let me know.

nextJSTypeScript/Dockerfile
microhobby: ok, I got it. Yeah, on my changes I turned back to `torizon/debian`, later we can set up the Chromium to run from docker-compose.yml

escherstair:
> @escherstair I pushed some changes, I'm using the `patch-package` to patch some issues from NextJS when trying to use the node `--inspect`. Also, now we are trying to attach to all the 3 processes using a compound. So, when you try to use it, select the target `Next.js Local`:
> Code_WRiAWwhiMm.mp4

Thanks @microhobby 
I see that you added `"postinstall"` script that calls `"pacth-package"`, but I'm not sure when it's called.
In your video I see the following messages:
immagine

When I try to do the same I have different messages
immagine

I see `"next dev"` instead of `NODE_OPTIONS=...` and no debugger attached.
I think I missed some steps.
I also tried to delete `node_modules` folder, but the beahvior is the same.

escherstair:
> Obs: I'm still not satisfied with the solution, for NextJS to run correctly it seems to need to run `npm install` on the remote target, so NPM gets the right libraries for armv8. But it takes a while each time I have to run the debug session. If you have ideas to improve this, let me know.

Not sure if this can work, but what happens if tjhe following steps
```
COPY package.json package-lock.json ./
RUN  npm install
```
are added to the `Dockerfile.debug`?
In this way the steps should be handled into the docker cache.
Not sure if the cache is invalidated when `package.json` changes.

microhobby:
> > @escherstair I pushed some changes, I'm using the `patch-package` to patch some issues from NextJS when trying to use the node `--inspect`. Also, now we are trying to attach to all the 3 processes using a compound. So, when you try to use it, select the target `Next.js Local`:

I only updated the `armv8`, are you trying to deploy in a `armv7` board?

microhobby:
> > Obs: I'm still not satisfied with the solution, for NextJS to run correctly it seems to need to run `npm install` on the remote target, so NPM gets the right libraries for armv8. But it takes a while each time I have to run the debug session. If you have ideas to improve this, let me know.
> 
> Not sure if this can work, but what happens if tjhe following steps
> 
> ```
> COPY package.json package-lock.json ./
> RUN  npm install
> ```
> 
> are added to the `Dockerfile.debug`? In this way the steps should be handled into the docker cache. Not sure if the cache is invalidated when `package.json` changes.

Probably, but let's test to see if this can work, if this gain some speed up it would be good

escherstair:
> I only updated the `armv8`, are you trying to deploy in a `armv7` board?

No, I refer to "Next.js local"

microhobby:
> I see that you added "postinstall" script that calls "pacth-package", but I'm not sure when it's called.
In your video I see the following messages:

@escherstair `postinstall` is automatically ran right after `npm install`, maybe you should have to remove your `node_modules` folder and try again, to be sure that the patches will be applied.

Sorry for the delay, I'm back now

escherstair:
> maybe you should have to remove your `node_modules` folder and try again, to be sure that the patches will be applied.

I had already tried it, but it didn't help.

microhobby:
> > maybe you should have to remove your `node_modules` folder and try again, to be sure that the patches will be applied.
> 
> I had already tried it, but it didn't help.

this is weird. And if you try in the terminal:

```
npm run postinstall
```

What's happen?

microhobby:
@escherstair some news about this?

escherstair:
@microhobby not yet.
I asked the help of a colleague to investigate, since I'm busy in another project.

microhobby:
Hey @escherstair let me know if you have plans to continue the contributions here. If not, we will close this MR, thanks in advance.

escherstair:
Hi @microhobby ,
yes, I and my colleague @dellorogiulio will contribute to this.
We need some more time to finalize our contribution.
Thanks for your patience.

dellorogiulio:
Hi @microhobby,
I'm working on the NextJS template for TC. 
However, recently NextJS changes its project structure (what they call [src directory](https://nextjs.org/docs/pages/building-your-application/configuring/src-directory). 
Therefore, generally these types of applications are bootstrapped via specific script (in this case, the [create next-app](https://nextjs.org/docs/pages/api-reference/create-next-app)). How can we handle this choice inside a template?

thanks!

microhobby:
Hey @dellorogiulio we should choose one of them and then set it as the recommended way, since we should have the tasks ready. So, stay free to choose the recommended way.

dellorogiulio:
@microhobby  I create a [new PR](https://github.com/toradex/vscode-torizon-templates/pull/127), once that is merged, you can close this one.
Thanks!

Issue #63, opened 2023-05-23
Typo in Dockerfile.debug for nodeJSTypeScript sample


escherstair:
In file `Dockerfile.debug` for nodeJSTypeScript sample ther is the following typo that refers to .NET Uno templates
https://github.com/toradex/vscode-torizon-templates/blob/1256dec394eaf04ae675d1499fd04c2cc159c571/nodeJSTypeScript/Dockerfile.debug#L31

I suggest to change it either to
```
# Deploy Step
```
as done in cppConsole
or to
```
# Deploy Step Node.js Typescript
```

microhobby:
It was fixed by 8f1483e41e875f23fe6f9c9d24064265efa785ca

PR #73, opened 2023-06-26
Fix typo in Dockerfile.debug for Node.js template

escherstair:
See issue  #63


microhobby:
LGTM! @escherstair thanks for the contribution!

PR #79, opened 2023-07-18
Fixes for RC 2.0.12

microhobby:


microhobby:
@tronical if possible please take a look at:

https://github.com/toradex/vscode-torizon-templates/pull/79/commits/4850fb67d61d5c2881e19ba341d97ba01234e23c

https://github.com/toradex/vscode-torizon-templates/pull/79/commits/6aa6ae6257e18825dff9d303bbd75e4b8533c0ec

We are using the pre-built Docker images from the community Containerfiles: https://github.com/commontorizon/Containerfiles

Let me know if this is ok on your side, thanks in advance.

tronical:
> @tronical if possible please take a look at:
> 
> [4850fb6](https://github.com/toradex/vscode-torizon-templates/commit/4850fb67d61d5c2881e19ba341d97ba01234e23c)
> 
> [6aa6ae6](https://github.com/toradex/vscode-torizon-templates/commit/6aa6ae6257e18825dff9d303bbd75e4b8533c0ec)
> 
> We are using the pre-built Docker images from the community Containerfiles: https://github.com/commontorizon/Containerfiles
> 
> Let me know if this is ok on your side, thanks in advance.

Thanks for making these changes to improve build times.

Overall this looks good to me.

One detail I noticed is that for the rustSlint SDK configuration, the slint-sdk-XX container is used, which contains Slint compiled for C++ in /usr - as well as the whole cmake/ninja setup. Similarly, the debug image for Rust will contain libslint_cpp.so. I /think/ it should not cause any conflict, but it's something to keep an eye on maybe.

One thing we do need to sort out though is the update process.

I can think of three scenarios:

1. You update base-version say from 3.0.9-bookworm to 3.10 (I'm making the version up :-). In that case you rebuild the slint-sdk/slint-base images, push them, and then update vscode-torizon-templates. Easy.
2. We make a new minor or patch release of Slint. We update the `release/1` branch and upload a newer Slint to crates.io.
  a. For C++, a local build (without docker) will pick up the latest 1.x release through [this](https://github.com/toradex/vscode-torizon-templates/blob/1256dec394eaf04ae675d1499fd04c2cc159c571/cppSlint/CMakeLists.txt#L13) reference. However builds for Torizon depend on slint-sdk/slint-base being updated. The image version referenced in Dockerfile however does not include a Slint version.
    b. For Rust, a local build (without docker) as well as a build for Torizon will pick up the latest minor release (unless the version is pinned in the application's `Cargo.toml`.).
3. A new Rust version is released. The slint-sdk/slint-base images should be updated, but existing users won't be able to benefit from updates, and they can't update to a newer version of Rust themselves because they're using slint-sdk/slint-base.

I think the scenario 2.a and 3. are problematic for users. Maybe they want to use a newer version of Slint (it's still a young product, so frequent updates are likely), but they can't. Similarly, Rust is still moving fast (while maintaining compatibility), but users can't update unless they re-create the docker files from the vs code extension.

Those were primary motivations for me for the full-fledged docker file, that is slower to build initially, but in the long term gives more power to users because they can make changes.

One way out would be to extend the version of the slint-sdk and slint-base images to include the slint version and the rust version. It doesn't mean that every combination needs to be built, latest slint release in combination with the rust version that's current at that point would suffice. Combine that with a comment in the docker file that tells users: Hey, if you want to update, this is what you need to change, and perhaps we'd have a flexible solution that doesn't compromise on the speed-up of providing pre-built images.

We can add an item on our TODO list (Slint) to notify you to rebuild the slint-sdk/slint-base images when we make a new release, but you retain control over what Torizon base image to choose, etc.

What do you think?

(cc @ogoffart - Olivier, I can fill you in on the current and the propose mechanism if you're interested)

tronical:
Ok, a brief chat later, suggestion: Use ${BASE_VERSION}-${SLINT_VERSION} as version for slint-sdk/slint-base and [here](https://github.com/commontorizon/Containerfiles/blob/6f730f95cd32e574d0bbe6d92e759aa2a7fe1763/slint-sdk/Containerfile#L87) select the corresponding vX.Y.Z tag instead of the branch. Then add a comment in this repo's Docker files that this selects the slint version and maybe a link to docker hub where people can see what versions are available. Would that make sense?

microhobby:
> We can add an item on our TODO list (Slint) to notify you to rebuild the slint-sdk/slint-base images when we make a new release, but you retain control over what Torizon base image to choose, etc.
> 
> What do you think?

@tronical I think that would be great to have this notification. As we have our template updater mechanism, I hope that users who want to use the newest versions will be able to do so easily whenever we have a new version of the images.


> Ok, a brief chat later, suggestion: Use ${BASE_VERSION}-${SLINT_VERSION} as version for slint-sdk/slint-base and [here](https://github.com/commontorizon/Containerfiles/blob/6f730f95cd32e574d0bbe6d92e759aa2a7fe1763/slint-sdk/Containerfile#L87) select the corresponding vX.Y.Z tag instead of the branch. Then add a comment in this repo's Docker files that this selects the slint version and maybe a link to docker hub where people can see what versions are available. Would that make sense?

Totally, I will work on these changes.

microhobby:
@tronical done! Let me know if I miss something, thanks

tronical:
Looks good - thanks!

PR #81, opened 2023-07-28
cConsole: add C console template

hiagofranco:
Template based on C++ template.


hiagofranco:
@microhobby: I wasn't sure if I should create a new ssh key and add it to the template, please correct me if I'm wrong.

microhobby:
Hey @hiagofranco, this is something that we should improve in the future, because now all templates has the same keys, that are duplicated between then. So, for now to create new templates just copy the keys and it's done.

microhobby:
Hey @hiagofranco LGTM! Thanks for the contribution!

PR #83, opened 2023-08-07
Hot Fixes for 2.1.0

microhobby:
Check the commits please

### DIFF
--- a/tcb/.conf/update.json
+++ b/tcb/.conf/update.json
@@ -1,2 +1,10 @@
 [
+    {
+        "source": ".github/workflows/build-torizoncore.yaml",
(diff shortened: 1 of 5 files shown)


tcb/.conf/update.json
```diff
@@ -1,2 +1,10 @@
 [
+    {
```
andreriesco, line 2: If the tcb template does not have this files, should it be here?

microhobby: Is not that it does not have it, but is not the application one. These are the specific TCB ones, they are different.

andreriesco:
LGTM

PR #91, opened 2023-08-18
tasks: Python: Upgrade pip on the host if it has version lower than 23

andreriesco:
Add a task to upgrade pip on the host if it has version lower than 23, when running a local debug.

### DIFF
--- a/python3Console/.vscode/tasks.json
+++ b/python3Console/.vscode/tasks.json
@@ -177,8 +177,26 @@
             "icon": {
                 "id": "layers",
                 "color": "terminal.ansiCyan"
+            },
+            "dependsOn": [
+                "host-upgrade-pip"
+            ]
+        },
+        {
+            "label": "host-upgrade-pip",
+            "detail": "",
+            "hide": true,
+            "type": "shell",
+            "options": {
+                "cwd": "${workspaceFolder}"
+            },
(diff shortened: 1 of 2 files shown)


python3Console/.vscode/tasks.json
```diff
@@ -177,8 +177,25 @@
...
                 "id": "layers",
                 "color": "terminal.ansiCyan"
+            },
+            "dependsOn": [
+                "host-upgrade-pip"
+            ]
+        },
+        {
+            "label": "host-upgrade-pip",
+            "detail": "hide",
```
microhobby: We are shipping now the new way, instead of the `"detail": "hide"` use `"hide": true`

python3Pyside2QML/.vscode/tasks.json
```diff
@@ -177,8 +177,25 @@
...
                 "id": "layers",
                 "color": "terminal.ansiCyan"
+            },
+            "dependsOn": [
+                "host-upgrade-pip"
+            ]
+        },
+        {
+            "label": "host-upgrade-pip",
+            "detail": "hide",
```
microhobby: We are shipping now the new way, instead of the `"detail": "hide"` use `"hide": true`

microhobby (changes requested):

microhobby:
LGTM

Issue #88, opened 2023-08-16
Qt CPP Template with wrong gnu++14 cppStandart


microhobby:
The `cppStandart` should be `gnu++17`

https://github.com/toradex/vscode-torizon-templates/blob/bookworm/cppQML/.vscode/c_cpp_properties.json#L12C38-L12C38

microhobby:
This was fixed on 0fbade56eb1762c75295fe05f6290e35461f8011

PR #93, opened 2023-08-18
cppQML: c_cpp_properties: Update cppStandard to gnu++17

microhobby:
Qt 6 uses C++17 as default, so update the cppStandard to gnu++17


microhobby:
Fixes #88

andreriesco:
LGTM

PR #94, opened 2023-08-21
Zig Template: zigConsole

kassane:
Hi @microhobby ,

website: https://ziglang.org

Template based on v0.11.0 [release-notes](https://ziglang.org/download/0.11.0/release-notes.html) - uses LLVM 16.0.6.

**Glossary:**

- `build.zig`: build-system written on zig, makefile like.
- `build.zig.zon`: dependencies manager, cargo.toml like.
- `zig`: toolchain
- `ziglang`: language

#### About:

Ziglang is a new language having the purpose to be an alternative to C, not as replacement only, also as complement (easy ffi - better than using C code in C++). Zig can guarantee that C code runs with clang rules and not clang++.

```bash
$> zig version
0.11.0
$> zig cc --version
clang version 16.0.6 (https://github.com/ziglang/zig-bootstrap 1dda86241204c4649f668d46b6a37feed707c7b4)
Target: x86_64-unknown-linux-musl
Thread model: posix
InstalledDir: /usr/bin
```

#### Current Status

Today, zig has llvm-backend to build C and C++ projects (`clang`/`clang++`) in which it manages porting some `libC` for easy cross-compilation ( CI added by me is proof of that), so it is first class support. Except for BSDs with incomplete support and MSVC because of MS license.

Zig libc: https://github.com/ziglang/zig/tree/master/lib/libc **(first class support = portability)**

Zig can also move between `glibc` versions. It makes it possible to choose the version of `glibc` in the target build.
https://github.com/ziglang/glibc-abi-tool

For C++ exclusively, zig has support `llvm-libcxx` + `llvm-libcxxabi` + `llvm-libunwind` + custom `libcompiler-rt` (rewritten on zig) that are statically linked in the binaries.

https://github.com/ziglang/zig/tree/master/lib/libcxx
https://github.com/ziglang/zig/tree/master/lib/libcxxabi
https://github.com/ziglang/zig/tree/master/lib/libunwind
https://github.com/ziglang/zig/tree/master/lib/compiler_rt

I'm a C++ developer currently porting some libraries which will allow me to use `zig-pkg` (package manager). Something equivalent to `fetchContent`(`cmake`), that is, for non-header-only libraries would need that dependencies have `cmake` to be built and linked to the main project. The same applies to `zig-pkg` which needs to push includes and libs built before linking to the main project.

Summarizing, `zig c++` build our app statically running `libcxx`+`libunwind` on almost all platforms under the same condition (C++ABI unified).

#### Zig is planning to not depend from LLVM. How will C++ support work?

Yes, however only version 1.0 that maybe will be released in mid 2025!

Explained here: https://youtu.be/AqDdWEiSwMM?t=2435

The zig project has always wanted to become a self-hosted toolchain making LLVM support optional, while benefiting from all the targets of this tool. However, many bugs inhibit further development of the main project (zig), which has fewer compiler developers.

- https://github.com/ziglang/zig/issues/16270#issuecomment-1616115039 (**rationale about C++ and obj-C++ support**)

[Andrew Kelley](https://github.com/andrewrk) has plans for zig to become not only a language and toolchain focused on the language itself, but also a new backend compiler that competes with LLVM and GCC keeping compatibility.

**Roadmap:** old diagram (v0.9.0)

- [How works zig `std.log`](https://gist.github.com/kassane/a81d1ae2fa2e8c656b91afee8b949426)


microhobby:
awesome! Let's review it, thanks for the contribution

.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
+{
+    // Use o IntelliSense para saber mais sobre os atributos possíveis.
+    // Focalizar para exibir as descrições dos atributos existentes.
+    // Para obter mais informações, acesse: https://go.microsoft.com/fwlink/?linkid=830387
```
microhobby: Could you please change here to English?

zigConsole/.conf/deps.json
```diff
@@ -0,0 +1,8 @@
+{
+    "packages": [
+        "openssh-client",
+        "sshpass",
+        "zig",
```
microhobby: This package is not presented in the default Debian source. There is a way to add the sources to get the zig packages?

microhobby (changes requested):

zigConsole/.conf/deps.json
microhobby: I saw that there is this one: https://github.com/dryzig/zig-debian/releases
Maybe we should add a task to download it and install it. What do you think?

kassane: > I saw that there is this one: https://github.com/dryzig/zig-debian/releases Maybe we should add a task to download it and install it. What do you think?

https://repology.org/project/zig/versions

`zig-debian` outdated.

Alternative: https://github.com/marler8997/zigup
or https://pypi.org/project/ziglang/ `python -m pip ziglang`

microhobby:
@kassane also is needed to add the template here: https://github.com/toradex/vscode-torizon-templates/blob/dev/templates.json

The VS Code extension uses these metadata, and lists the templates when creating new projects.

kassane:
> @kassane also is needed to add the template here: https://github.com/toradex/vscode-torizon-templates/blob/dev/templates.json

Ok.

microhobby:
Hey @kassane I took the liberty of switching to `dev` and running a rebase. I didn't see it on the first iteration, but everything new first goes into the `dev` branch.

microhobby:
I'm checking here, we are missing the `.vscode/` files, like the `settings.json`, `extension.json` and etc... Take a look in one of the other templates, these files are needed to install the right extensions and debugger.

kassane:
> I'm checking here, we are missing the `.vscode/` files, like the `settings.json`, `extension.json` and etc... Take a look in one of the other templates, these files are needed to install the right extensions and debugger.

Ok. (also rebased to `toradex/dev`)

kassane:
@microhobby ,

After this PR, started another project since there was no docker container with latest zig (no bootstraping). So, I started running some tests.

I even suggest you see here are some of my tests:
- https://github.com/kassane/docker-zig/issues/1

For the main purpose of this PR it is not impaired because (cited in the issue above) ziglang projects do not have any problems during build, as they do not need libc and Debian/Alpine containers works normally.

But, still looking into the problem with libunwind in the (busybox) container.

I'll do more testing with the other containers (Debian and Alpine).

microhobby:
Hey @kassane I will remove the milestone for the `v2.2.0` because it's scheduled for the next week. So, we will have no time to review it and handle the fixes until `v2.2.0`.

kassane:
New update with CI runner (rebased). @microhobby 

--- edit

Sometimes CI makes zig runs longer than totalized rust.

microhobby:
hey @kassane thanks for the work, I was on vacation, but I'm back. I will take a look this week

zigConsole/.vscode/launch.json
```diff
@@ -0,0 +1,211 @@
...
+                    "text": "-enable-pretty-printing",
+                    "ignoreFailures": true
+                },
+                {
+                    "description":  "Set Disassembly Flavor to Intel",
+                    "text": "-gdb-set disassembly-flavor intel",
+                    "ignoreFailures": true
+                }
+            ],
+            "preLaunchTask": "build-debug-amd64-local"
```
microhobby: this tasks appears to not be present, I check and there is the `build-debug-amd64`

zigConsole/.vscode/launch.json
```diff
@@ -0,0 +1,211 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Local AMD64",
+            "type": "cppdbg",
+            "request": "launch",
+            "program": "${workspaceFolder}/zig-deploy/x86_64-linux-gnu/debug/bin/__change__",
```
microhobby: when using the `build-debug-amd64` it's appears to build it correctly, but then it's create the `${workspaceFolder}/zig-out/bin/apptest` instead. So, changing here appears to work.

This is wat you are expecting, or the task `build-debug-amd64` should be changed to output to a another folder structure?

microhobby (changes requested):

zigConsole/.vscode/launch.json
kassane: Based on C and C++ solution

https://github.com/toradex/vscode-torizon-templates/blob/bookworm/cConsole/.vscode/launch.json#L28

zigConsole/Dockerfile.sdk
```diff
@@ -0,0 +1,34 @@
...
+# BUILD ------------------------------------------------------------------------
+FROM torizon/debian-cross-toolchain-${IMAGE_ARCH}:${CROSS_SDK_BASE_TAG}
+
+# __deps__
+RUN apt-get -q -y update && \
+    apt-get -q -y install \
+    # ADD YOUR PACKAGES HERE
+    && \
+    apt-get clean && apt-get autoremove && \
+    rm -rf /var/lib/apt/lists/*
```
microhobby, line 8: here you are using the `torizon/debian-cross-toolchain-<arch>` but are not installing zig, this image has only the build essentials for the architecture chosen: https://github.com/toradex/torizon-containers/blob/bookworm/debian-cross-toolchains/base/Dockerfile

if you try to run the `build-debug-arm64` for example, this will error with:

```
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "zig": executable file not found in $PATH: unknown.
```

microhobby (changes requested):

zigConsole/Dockerfile.sdk
kassane: new container added on dockerfile sdk
https://hub.docker.com/r/kassany/bookworm-ziglang

kassane: update: new commit on `docker-zig` add multiples architectures on `bookworm-ziglang` image:
https://github.com/kassane/docker-zig/commit/414d2851564f7f822407618c6c0989c91cacf2e0

templates.json
```diff
@@ -458,6 +458,27 @@
...
                 "successDebugMessage": "Hello Torizon!"
             }
+        },
+        {
+            "img": "zig-console.png",
+            "folder": "zigConsole",
+            "description": "Zig Console Application",
+            "runtime": "no-libc",
+            "language": "Zig",
+            "debugSSHPort": "2320",
```
microhobby, line 468: here we are setting "2320" but in the docker-compose.yml using the "2230".
Please change the docker-compose.yml to use "2320" too.

zigConsole/docker-compose.yml
```diff
@@ -0,0 +1,15 @@
+version: "3.9"
+services:
+  __container__-debug:
+    build:
+      context: .
+      dockerfile: Dockerfile.debug
+    image: ${LOCAL_REGISTRY}:5002/__container__-debug:${TAG}
+    ports:
+      - 2230:2230
```
microhobby: here should we use the same described from the templates.json

zigConsole/.vscode/tasks.json
```diff
@@ -0,0 +1,190 @@
...
+                "-i",
+                "${workspaceFolder}/.conf/id_rsa",
+                "-o",
+                "UserKnownHostsFile=/dev/null",
+                "-o",
+                "StrictHostKeyChecking=no",
+                "-P",
+                "${config:torizon_debug_ssh_port}",
+                "-pr",
+                "${workspaceFolder}/zig-deploy/aarch64-linux-gnu/debug/bin",
```
microhobby: we should change all the "deploy-torizon-<arch>" to sync from "${workspaceFolder}/zig-out/bin"

zigConsole/.vscode/tasks.json
```diff
@@ -0,0 +1,190 @@
...
+            "args": [
+                "docker",
+                "run",
+                "--rm",
+                "-it",
+                "-v",
+                "${workspaceFolder}:/app",
+                "cross-toolchain-arm64-__container__",
+                "zig",
+                "build",
```
microhobby, line 48: since all the builds will be stored in the same zig-cache and zig-out I think that we can add here an argument:

```suggestion
                "build",
                "--cache-dir", "zig-aarch64",
```

so, we can maintain the caches in different folders and reuse

microhobby: please add this for all the "build-debug-<arch>"

microhobby (changes requested):

microhobby:
@kassane thanks for the great work, we are very close to the end!

kassane:
Nice! Thank you for your attention and patience.
If you have any questions, just ask!!

zigConsole/.vscode/tasks.json
kassane: Regardless of the zig-cache, it is important to note that it still needs the global-cache in `$HOME/.cache/zig`

```bash
 --cache-dir [path]           Override path to local Zig cache directory # (only your application)
  --global-cache-dir [path]    Override path to global Zig cache directory # (build - libcompiler-rt, libc, zig-pkg , libcxx, libunwind)
```

kassane: > please add this for all the "build-debug-"

Do you want me to separate the binaries copied to the zig-out?

For searching in the zig-cache it will depend on the hash generated. This would mean that if the build is done several times under the same target with some changes, `zig-cache` will retain the old objects and add new objects and final binaries. However, users don't know exactly which hash was generated by the latest build.
Currently, maybe delete `zig-cache` before next build or make random cache folders.

<details>
<summary>zig-cache tree-view</summary>

```bash
zig-cache
├── h
│   ├── 81193319b29ea4a53c08ab4ac0d3a45a.txt
│   ├── b0cd276c88bf14583d0013b7874ad23e.txt
│   ├── cec882ddf082978e4cb5f3e0fe70a406.txt
│   ├── ed5f09d49558e684a6b652479ec63204.txt
│   └── timestamp
├── o
│   ├── 08e7e72ed8f2779ad0951972c7b727a5
│   │   ├── build
│   │   └── build.o
│   ├── 0e1478d8fb3e5e346e185ea049bc7571
│   │   ├── build
│   │   └── build.o
│   ├── 10c923e78aed7516d398c945a87ab2e3
│   │   ├── __change__
│   │   └── __change__.o
│   ├── 18f574651c2ac3f911e2ce692db80334
│   │   ├── build
│   │   └── build.o
│   ├── 1a94247c0ec90f6600d5958f06f911b4
│   │   └── dependencies.zig
│   ├── 3d25ba929badfaa7fcc3d38af6f9532c
│   │   ├── build
│   │   └── build.o
│   ├── 6f1f4d594305e608a2e3197e88d44c98
│   │   ├── __change__
│   │   └── __change__.o
│   ├── 73b91981e7c845ffbf9d73aba9b6973f
│   │   └── builtin.zig
│   ├── a758f8b37a973e406eeeefce72f97572
│   │   ├── __change__
│   │   └── __change__.o
│   ├── ccea54b0fc2c66e9206d70b2c62f9cc2
│   │   └── builtin.zig
│   ├── d996a26408e21cd94fabc66317c94674
│   │   └── builtin.zig
│   └── dc680198a4666a4baebb4e11152f4998
│       ├── build
│       └── build.o
├── tmp
│   └── 6b70b395b0110e23
└── z
    ├── 006c7bbcd79f49487376e2dd5c87e566
    ├── 36e060f54e0be4056ddf2a5550928c01
    ├── 506a4f47720b6f4d9a96ceb0098de141
    └── f376aea89cf12748e9803ad17ae77daa

18 directories, 29 files
```
</details>

kassane: The original proposal had been to add a custom path separating the final binaries copied from the `zig-cache`.
However, it was reversed since I had an issue with docker where the binary wasn't recognized nor found by the path specified.

**old:** replacing `zig-out/` to `zig-deploy/{{triple-arch}}/{{optimizerMode}}`
https://github.com/toradex/vscode-torizon-templates/blob/08eaf344e7f3145138daf22eaa7270fd0b16a661/zigConsole/Dockerfile#L103-L104

```diff
diff --git a/zigConsole/build.zig b/zigConsole/build.zig
index c7a870e..b538f2d 100644
--- a/zigConsole/build.zig
+++ b/zigConsole/build.zig
@@ -32,4 +32,30 @@ pub fn build(b: *std.Build) void {
 
     // copy binary from zig-cache to zig-out/bin [default]
     b.installArtifact(binary);
+
+    // overwrite default output
+    b.resolveInstallPrefix(b.fmt("zig-deploy/{s}/{s}", .{ target.linuxTriple(b.allocator) catch unreachable, @tagName(optimize) }), .{});
```

kassane: ### Suggestion

Change:
```zig
    // overwrite default output
    b.resolveInstallPrefix(b.fmt("zig-out/{s}/{s}", .{ @tagName(target.getCpuArch()), @tagName(optimize) }), .{});
    // tagName convert tagged Enum fields to string ([]const u8)
```

**Output:**
```bash
zig-out
├── aarch64
│   ├── Debug
│   │   └── bin
│   │       └── __change__
│   └── ReleaseSafe
│       └── bin
│           └── __change__
└── x86_64
    ├── ReleaseFast
    │   └── bin
    │       └── __change__
    └── ReleaseSafe
        └── bin
            └── __change__
```

**Reference:**
| Mode | Runtime Safety | Optimizations |
|---------------|----------------|---------------|
| Debug         | Yes            | No            |
| ReleaseSafe  | Yes            | Yes, Speed    |
| ReleaseSmall | No             | Yes, Size     |
| ReleaseFast  | No             | Yes, Speed    |

microhobby: sounds good to me, go ahead!

kassane: ok. New commit

microhobby:
@kassane this looks good, but there is another detail. We need a way to add the pretty printer to the Debug session: https://github.com/ziglang/zig/blob/master/tools/zig_gdb_pretty_printers.py

Without this the variables look weird on the VS Code inspector

kassane:
> @kassane this looks good, but there is another detail. We need a way to add the pretty printer to the Debug session: https://github.com/ziglang/zig/blob/master/tools/zig_gdb_pretty_printers.py
> 
> Without this the variables look weird on the VS Code inspector

Ok. Need more scripts `*.py`? Zig v0.11.0 or master are stage2 only.

microhobby:
> Ok. Need more scripts *.py? Zig v0.11.0 or master are stage2 only.

To be honest I have no idea. I tried the configuration locally and on the remote but these do not work as expected. Do you have these working on your machine?

kassane:
> Without this the variables look weird on the VS Code inspector

@microhobby ,

Pretty printer not work in all and on zig is experimental.

I put scripts in my configuration and didn't see much difference.
The example added, a log print, doesn't show much. However, when adding a variable with the name torizon, the inspector gives information.

**cppdbg** w/ gdb

<details>
<summary>preview - print</summary>

image
</details>

**LLDB**

<details>
<summary>preview - print</summary>

image
</details>

**References**
- https://github.com/ziglang/zig/issues/7283
- https://github.com/ziglang/zig/issues/7361
- https://github.com/ziglang/zig/issues/7505
- https://github.com/ziglang/zig/issues/15095
- https://github.com/ziglang/zig/issues/13095

microhobby:
Ok, so I think that we have the same problem from the Rust templates #96, the LLDB works in a way that shows the variables in a better presentation but also we should figure out how to use it in the remote environment.

So, I think that we can drop the ebe3cd5ea20b3157ec34e1a2cef27f014460f9cd and 69f5341c4059a659033c349816a158e009dcbe2c and it's done

kassane:
> So, I think that we can drop the [ebe3cd5](https://github.com/toradex/vscode-torizon-templates/commit/ebe3cd5ea20b3157ec34e1a2cef27f014460f9cd) and [69f5341](https://github.com/toradex/vscode-torizon-templates/commit/69f5341c4059a659033c349816a158e009dcbe2c) and it's done

dropped commits

microhobby:
LGTM

PR #106, opened 2023-08-31
tasks: common: Fix usage of the env variables for tcb-platform-publish

### LINKED ISSUE #90: `tcb-platform-publish` work on the CI pipeline but does not run correctly locally
This was reported on the Toradex community:
https://community.toradex.com/t/ota-issues-from-torizon-platform-and-github-actions/20077/31

We need to check if the set of environment variables can be used in the same task that is using it. We are defining the env and using it on the same `env` task, maybe this is not supported. The way that the CLI interprets it works, because it sets in the same context.

https://github.com/toradex/vscode-torizon-templates/blob/34e86493492453eb2559a1ff6adbed0cd979013b/assets/tasks/common.json#L226

microhobby:
fixes #90

### DIFF
--- a/assets/tasks/common.json
+++ b/assets/tasks/common.json
@@ -204,7 +204,7 @@
                 "env": {
                     "DOCKER_PSSWD": "${command:docker_password}",
                     "DOCKER_LOGIN": "${command:docker_login}",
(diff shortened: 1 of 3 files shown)


andreriesco:
LGTM

PR #109, opened 2023-09-01
`torizoncore-dev` Series

microhobby:
Please check the commits


scripts/zygote.ps1
```diff
@@ -35,6 +35,9 @@ function help() {
     Write-Host " ➡️ new"
     Write-Host "`t Create a new project from a template using TUI"
```
andreriesco, line 36: Here should be TUI or GUI?

microhobby: yeah, should be TUI because is a Terminal Graphical Interface, is not Graphical at end

andreriesco:
LGTM

PR #111, opened 2023-09-11
templates: Add maintainence policy

### LINKED ISSUE #31: What is the maintenance and deprecation policy for the community templates?
I've read https://github.com/toradex/vscode-torizon-templates/blob/main/CONTRIBUTING.md and didn't find a note on the maintenance policy for the templates.

With time, they may break so community contributors should be aware they are expected to update the templates, otherwise, we will end up with many non-working ones.

Also, we should have a deprecation policy for when a community template becomes unmaintained for any reason, for example, the original maintainer doesn't want to keep it anymore or ourselves.

It might be good to think of a badge to add on each template on https://github.com/toradex/vscode-torizon-templates#community-supported-templates, maybe with the "last updated" date, or something that can indicate to users the health of the community templates. Most users will not come to this GitHub repo, so this info should also be displayed in VS Code when creating a new project from a template.

microhobby:
fixes #31

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -155,3 +155,17 @@ For ApolloX project templates we are using [Github platform](https://docs.github
     ```
 
 - Create a pull request to the `dev` branch of https://github.com/toradex/vscode-torizon-templates
+
+## Maintainence Policy
+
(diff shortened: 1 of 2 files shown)


andreriesco:
LGTM

PR #122, opened 2023-09-25
tasks: Add the tasks to build and run locally the production images

andreriesco:
This commit adds the tasks to:
. Build the application based in the  production Dockerfile . Push it to the local registry (localhost:5002)
. Pull it on the board from the local registry
. And run it on the board, showing the result in the VSCode terminal

### DIFF
--- a/assets/tasks/common.json
+++ b/assets/tasks/common.json
@@ -1659,7 +1659,7 @@
             ],
             "problemMatcher": "$msCompile",
             "icon": {
-                "id": "flame",
+                "id": "run",
                 "color": "terminal.ansiYellow"
             }
         },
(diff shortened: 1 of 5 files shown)


assets/tasks/common.json
```diff
@@ -1518,80 +1438,327 @@
...
+                "copy-docker-compose",
+                "pre-cleanup-arm64",
+                "build-container-torizon-release-arm64",
+                "push-container-torizon-release-arm64",
+                "pull-container-torizon-release-arm64",
+                "wait-a-bit"
+            ],
+            "problemMatcher": "$msCompile",
+            "icon": {
+                "id": "flame",
```
microhobby: @andreriesco please check if there is other icon that could be used here, should be more meaningful for the task

microhobby:
LGTM

PR #125, opened 2023-10-16
   cppQML: Add the hardware acceleration support for the Qt template

andreriesco:
Add the hardware acceleration support for the Qt template, by
    using the newly release Torizon Qt6 containers (torizon/qt6-wayland and
    torizon/qt6-wayland-vivante).
    Also change the build tool from qmake6 to cmake, because at the moment
    the "quick" module when added through qmake6 (in the .pro file),
    adds the requirement of libGLX on the Makefile, which is not available
    for the iMX8 modules.
    Therefore, before it gets fixed, to use qmake6 you have remove the
    /usr/lib/aarch64-linux-gnu/libGLX.so from the LIBS of the Makefile
    generated by qmake6 before running "make". Because of that, the
    recommended way is to use CMake, the default of this new
    hardware accelerated Qt6 template.
    
    Also, minor fixes on some tasks and on README.md.
    

microhobby:
@andreriesco when running the `build-debug-arm` I'm getting the follow error:

```
Executing task: DOCKER_HOST= docker run --rm -it -v /home/castello/tmp/vscode/qttesting:/app cross-toolchain-arm-qttesting cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_COMPILER=arm-linux-gnueabihf-g++ -DCMAKE_C_COMPILER=arm-linux-gnueabihf-gcc -Bbuild-armhf 

WARNING: The requested image's platform (linux/arm64/v8) does not match the detected host platform (linux/amd64/v3) and no specific platform was requested
-- The CXX compiler identification is unknown
-- The C compiler identification is unknown
CMake Error at CMakeLists.txt:2 (project):
  The CMAKE_CXX_COMPILER:

    arm-linux-gnueabihf-g++

  is not a full path and was not found in the PATH.

  Tell CMake where to find the compiler by setting either the environment
  variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
  to the compiler, or to the compiler name if it is in the PATH.


CMake Error at CMakeLists.txt:2 (project):
  The CMAKE_C_COMPILER:

    arm-linux-gnueabihf-gcc

  is not a full path and was not found in the PATH.

  Tell CMake where to find the compiler by setting either the environment
  variable "CC" or the CMake cache entry CMAKE_C_COMPILER to the full path to
  the compiler, or to the compiler name if it is in the PATH.
```

assets/tasks/common.json
```diff
@@ -943,7 +949,9 @@
                 "-t",
                 "cross-toolchain-riscv64-__container__",
                 "--build-arg",
-                "IMAGE_ARCH=riscv64"
+                "IMAGE_ARCH=riscv64",
+                "--build-arg",
+                "GPU=${config:torizon_gpu}"
```
microhobby: @andreriesco please drop this commit, these are not needed for the SDK containers, the linker will link the binary with the generic lib, and the specific GPU ones will be patched on the runtime container

microhobby (changes requested):

microhobby:
Fix #51

assets/tasks/common.json
andreriesco: I think it makes sense to keep just the arm64 one, as it uses the "-vivante" containers, which may have differences from the upstream even when compiling (like the libGLX support for example)

microhobby:
LGTM!

PR #128, opened 2023-10-27
cppQML: vscode: tasks: Add open-in-qtcreator for each architecture

microhobby:
Now on the start of the GDB server we open a new instance of the QtCreator with the correct kit for the architecture.

### DIFF
--- a/cppQML/.vscode/tasks.json
+++ b/cppQML/.vscode/tasks.json
@@ -24,6 +24,36 @@
         },
         {
             "label": "open-in-qt-creator",
+            "detail": "This task first builds the project and then opens it in Qt Creator (for local development/debug).\nNeed to have the Qt Creator installed on your system.",
+            "command": "qtcreator",
+            "type": "process",
+            "options": {
+                "env": {
+                    "QT_QPA_PLATFORM": "xcb"
+                }
+            },
+            "args": [
+                "${workspaceFolder}",
+                "-settingspath",
+                "${workspaceFolder}/.qt",
+                "-theme",
+                "dark"
+            ],
+            "problemMatcher": [
+                "$gcc"
+            ],
+            "icon": {
+                "id": "tools",
+                "color": "terminal.ansiGreen"
+            },
+            "dependsOrder": "sequence",
+            "dependsOn": [
+                "build-debug-amd64-local",
+                "update-qt-ini-amd64-local"
+            ]
+        },
+        {
+            "label": "open-in-qt-creator-debug-arm64",
             "detail": "This task first builds the project and then opens it in Qt Creator.\nNeed to have the Qt Creator installed on your system.",
             "command": "qtcreator",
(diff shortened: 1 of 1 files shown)


cppQML/.vscode/tasks.json
```diff
@@ -120,10 +137,12 @@
             ],
             "dependsOrder": "sequence",
             "dependsOn": [
+                "kill-qt-creator",
                 "update-qt-ini-arm64",
                 "kill-redirect-qml-debugger",
                 "deploy-torizon-arm64",
-                "redirect-qml-debugger"
+                "redirect-qml-debugger",
+                "open-in-qt-creator"
```
andreriesco: I think it is better to put the the `open-in-qt-creator` before the `deploy-torizon-arm64`, because it will take much less time to reopen, it will look much more like a "reload-window"

andreriesco: Or like this, as `deploy-torizon-arm64` validates that the arch is correct:
                "kill-redirect-qml-debugger",
                "deploy-torizon-arm64",
                 "kill-qt-creator",
                "update-qt-ini-arm64",
                "open-in-qt-creator",
                "redirect-qml-debugger"

microhobby: yeah, make sense

microhobby:
@andreriesco take a look on the modifications

cppQML/.vscode/tasks.json
```diff
@@ -120,9 +210,10 @@
             ],
             "dependsOrder": "sequence",
             "dependsOn": [
-                "update-qt-ini-arm64",
                 "kill-redirect-qml-debugger",
                 "deploy-torizon-arm64",
+                "update-qt-ini-arm64",
```
andreriesco: Will a new Qt Creator instance be opened every time this task is executed?

microhobby: @andreriesco Yes, this is the best and more quick way to do it, also the most safe because does not cancel any work unsaved from the user without their consent.

andreriesco: Actually I think we can include the `update-qt-ini-arm64` on the `dependsOn` of `open-in-qt-creator-debug-arm64`, and remove both of this from this `dependsOn`.
In that way we will not be closing the Qt Creator from the user and it will not be necessary to open many instances of Qt Creator. The user experience will be -> open Qt Creator for the desired arch -> run this task -> make modifications on the project in this Qt Creator -> run this task again -> make modifications on the project in this Qt Creator -> run this task again -> ...

andreriesco: Just like how it is today, but opening Qt Creator with the right settings for the desired arch

microhobby: hmm, ok make sense, let me change it

andreriesco:
LGTM

PR #131, opened 2023-11-08
cppQML: tasks: Remove build <arch> for open-in-qt-cretor-<arch>

microhobby:
For the remote debug on the Qt Creator we should use the local build and the local kit. If we use the remote build we will have to configure the kit for each of the <arch>. Also the path should be the same as the local build instead of the container ones.


microhobby:
LGT myself

PR #135, opened 2023-11-17
Release fixes and bumps

andreriesco:


dotnetUno/__change__.Skia.Gtk/__change__.Skia.Gtk.csproj
```diff
@@ -19,11 +19,11 @@
...
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Logging" Version="5.0.0" />
-    <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="5.0.0" />
-    <PackageReference Include="Uno.UI.Skia.Gtk" Version="4.8.33" />
-    <PackageReference Include="Uno.UI.RemoteControl" Version="4.8.33" Condition="'$(Configuration)'=='Debug'" />
-    <PackageReference Include="Uno.UI.Adapter.Microsoft.Extensions.Logging" Version="4.8.33" />
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="7.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="7.0.0" />
+    <PackageReference Include="Uno.UI.Skia.Gtk" Version="4.10.13" />
```
microhobby, line 24: this should be `4.10.23` not?
https://www.nuget.org/packages/Uno.UI.Skia.Gtk/

microhobby (changes requested):

assets/tasks/common.json
```diff
@@ -1648,6 +1563,8 @@
                 "TAG=arm64",
                 "GPU=${config:torizon_gpu}",
                 "docker-compose",
+                "-p",
+                "torizon",
```
microhobby, line 1567: did you think that is better to pass `torizon` than leave it to create with the service name?

microhobby:
@andreriesco as I need these changes to make some tests this weekend I will approve and merge it as is. But let's discuss these points that I commented monday.

dotnetUno/__change__.Skia.Gtk/__change__.Skia.Gtk.csproj
andreriesco: Interesting, I got the version from here: [https://github.com/unoplatform/uno/releases](https://github.com/unoplatform/uno/releases). Do you know if there is any specific reason for them not to publish every version on Gihub?

assets/tasks/common.json
andreriesco: This one is needed because `docker-compose down --remove-orphans` only stops containers with the same project name. So the debug container will not be stopped when trying to start the release container, and vice-versa, which generates problems like "This port is already allocated". 
This statement is also valid for trying to debug a new project after having debugged another one. This older one will not be stopped if it has a different project name.

PR #139, opened 2023-12-01
genericTemplate: Add a Generic Dockerfile Template (Without language)

andreriesco:
This is a special kind of generic template that is not made for any specific language and therefore does not have any debugging tasks, just the release ones (which uses the **Dockerfile** instead of the **Dockerfile.debug**).
For some examples of when this is useful, check the README.md file, on the genericTemplate/.doc folder.


assets/img/genericTemplate.png
microhobby, line 1: The contrast between this blue and the gray background is weird. Could you please change the Docker logo color, or add some background, I do not know, be creative 😃

genericTemplate/.vscode/launch.json
microhobby, line 1: I do not know if is a good idea to add a debug launch for this kind of project. Could you explain what is your idea behind this?

genericTemplate/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+
+
+# Command executed when starting the container
+CMD [ "echo", "Generic", "Dockerfile", "Template", "successfully", "executed" ]
+
+# This can be used to keep the container up indefinely, and than you can 
+# attach to it's shell by clicking on the Docker extension icon here on the
+# left tab here in VSCode, right clicking on the right container 
+# and selecting "Attach shell"
+# CMD ["sleep", "infinity"]
```
microhobby: also think that this is not needed, maybe add it to the documentation?

scripts/projectUpdater.ps1
```diff
@@ -279,16 +282,17 @@ Write-Host -ForegroundColor DarkGreen "✅ tasks.json"
 
 # TCB does not have the common application Docker files
 if ($templateName -ne "tcb") {
+    # DOCKERFILE.DEBUG:
+    # The generic project doesn't have a Dockerfile.debug
```
microhobby, line 286: why not? I have some ideas, like have a debug image that the user can apt-install stuff, change, and we have some command to get this changes and add it to the production image. What do you think?

microhobby (changes requested):

genericTemplate/.vscode/launch.json
andreriesco: The idea behind this is to print a message instructing the user to use the release task if he tries to run this with `F5` like the other templates, instead of letting VSCode print an error that will not help the user:
`This template has no debug, use the run-container-torizon-release-<arch> task (TASK RUNNER tab) to build the container and run it on the board`

genericTemplate/Dockerfile
andreriesco: Sure, I think it makes sense on the documentation

assets/img/genericTemplate.png
andreriesco: Sure

scripts/projectUpdater.ps1
andreriesco: The main idea was to simplify the usage of the template. In this specific template the user will likely have to customize the Dockerfile to his use case, and the Dockerfile.debug is harder to understand and less straightforward to customize. So, as there is no language to be debugged, the idea was to make it simpler and closer to a common Docker workflow the better for the user. Also, in many cases this will be used in a multi-container environment which complicates things even more

assets/img/genericTemplate.png
microhobby: have you changed the image?

scripts/projectUpdater.ps1
microhobby: so, this will not be added as multi-container compatible, right?

andreriesco: Actually most use cases will be used in a multi-container project I believe

assets/img/genericTemplate.png
andreriesco: No, not yet

genericTemplate/Dockerfile
andreriesco: Update it, please review the .doc/README.md

scripts/projectUpdater.ps1
andreriesco: Also for here, please review the README.md

andreriesco:
Closes: #140

genericTemplate/.vscode/tasks.json
```diff
@@ -0,0 +1,19 @@
+{
+    "version": "2.0.0",
+    "tasks": [
+        {
+            "label": "use-release-task-message",
+            "command": "echo  'This template has no debug, use the run-container-torizon-release-<arch> task (TASK RUNNER tab) to build the container and run it on the board'",
```
microhobby, line 6: should be good to also have this task returning some error code, so the debug session terminates as soon as possible. If not the debug session will be running until the timeout, I think that this could confuse the user.

microhobby (changes requested):

genericTemplate/.vscode/tasks.json
andreriesco: Actually the `exit` parameter here at `launch.json` is doing this:
`
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Use release task message",
            "type": "node",
            "request": "launch",
            "args": [
                "exit"
            ],
            "preLaunchTask": "use-release-task-message"
        }
    ]
}`

andreriesco: Found a nicer way to do it than how I was doing

andreriesco: Updated the template

microhobby:
LGTM

PR #141, opened 2023-12-05
cppConsole and cConsole: Fix clean tasks

andreriesco:
Fix clean tasks that were not cleaning the build directory, and add the build-makedir-<arch> task in the dependsOn of the build-debug-<arch> tasks.


microhobby:
LGTM

PR #142, opened 2023-12-12
Change SSH executables copy to Docker COPY and standardize app root

andreriesco:
. Changes the copy of executables from the SSH on the running
container to copy them with COPY in the Dockerfiles on the container
build.
. Standardizes between the SDK, the debug and the production the
APP_ROOT directory variable, to use a value defined on settings.json,
making it clear for the user where his project files are.
Does the same for the APP_EXECUTABLE file.
. Adds the information about how to deal with extra files (not added to
the running container automatically) on the documentation of each
template.


cmakeConsole/.vscode/tasks.json
```diff
@@ -35,7 +35,7 @@
                 "-DCMAKE_BUILD_TYPE=Debug",
                 "-DCMAKE_CXX_COMPILER=arm-linux-gnueabihf-g++",
                 "-DCMAKE_C_COMPILER=arm-linux-gnueabihf-gcc",
-                "-Bbuild-armhf"
+                "-Bbuild-arm"
```
microhobby, line 38: so, this is wrong also on the `bookworm` branch?

microhobby, line 38: so, this is wrong also on the `bookworm` branch?

microhobby:
@andreriesco still has one change that should be added. When doing `docker compose down` add the `--rmi local` argument. Because every time we do a new debug and there are changes, a new untagged image will be left behind, this can leave garbage in storage memory.

cmakeConsole/.vscode/tasks.json
andreriesco: It is not wrong actually, it is just that if we use `build-arm` instead of `build-armhf` we can pass the build directory like this on the `COPY build-${IMAGE_ARCH}/debug ${APP_ROOT}/app`

microhobby:
LGTM

PR #152, opened 2023-12-22
Add support for pushing to container registries different than Dockerhub

andreriesco:

### DIFF
--- a/assets/tasks/common.json
+++ b/assets/tasks/common.json
@@ -110,10 +110,11 @@
                 "-nop",
                 "${workspaceFolder}/.conf/createDockerComposeProduction.ps1",
                 "${workspaceFolder}",
-                "${command:docker_registry}",
+                "${command:docker_login}",
                 "${command:docker_tag}",
+                "${command:docker_registry}",
                 "__container__",
-                "${command:torizon_gpu}"
+                "${config:torizon_gpu}"
             ],
             "dependsOrder": "sequence",
             "icon": {
@@ -155,6 +156,7 @@
                 "cwd": "${workspaceFolder}",
                 "env": {
                     "DOCKER_REGISTRY": "${command:docker_registry}",
+                    "DOCKER_LOGIN": "${command:docker_login}",
                     "DOCKER_PSSWD": "${command:docker_password}",
                     "DOCKER_TAG": "${command:docker_tag}",
                     "TCB_CLIENTID": "${command:tcb.clientId}",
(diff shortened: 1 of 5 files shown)


scripts/createDockerComposeProduction.ps1
```diff
@@ -136,8 +141,8 @@ Write-Host -ForegroundColor DarkGreen "✅ Image rebuild and tagged"
 # push it
 Write-Host "Pushing it $dockerLogin/$($imageName):$tag ..."
 
-docker login --username $dockerLogin --password $psswd
-docker push $dockerLogin/$($imageName):$tag
+Write-Output "$psswd" | docker login $registry -u $dockerLogin --password-stdin
```
microhobby, line 151: this works?

microhobby:
@andreriesco also take a look at the TCB template, because some common stuff is not done there, and will possibly need an adjust. Let me know.

scripts/createDockerComposeProduction.ps1
andreriesco: Yes, I tested it on the Github registries (packages) and Gitlab registries, but also took a look at how it should be done for Azure and AWS.

andreriesco:
For tcb it was still missing. I fixed it and rebased the code, now it is working also for the `tcb-platform-publish` and `tcb-platform-publish-ostree` commands

assets/tasks/common.json
```diff
@@ -110,10 +110,11 @@
                 "-nop",
                 "${workspaceFolder}/.conf/createDockerComposeProduction.ps1",
                 "${workspaceFolder}",
-                "${command:docker_registry}",
+                "${command:docker_login}",
                 "${command:docker_tag}",
+                "${command:docker_registry}",
                 "__container__",
-                "${command:torizon_gpu}"
+                "${config:torizon_gpu}"
```
microhobby, line 117: Why? Changing this to `config:` will not show the input for the user 🤔

assets/tasks/common.json
```diff
@@ -236,10 +238,12 @@
                 "env": {
                     "DOCKER_PSSWD": "${command:docker_password}",
                     "DOCKER_LOGIN": "${command:docker_login}",
-                    "VSCODE_CMD": "--verbose platform push --credentials credentials.zip --package-name __change__ --package-version ${command:tcb.getNextPackageVersion} --login $DOCKER_LOGIN $DOCKER_PSSWD --canonicalize docker-compose.prod.yml"
+                    "DOCKER_REGISTRY": "${command:docker_registry}",
+                    "VSCODE_CMD": "--verbose platform push --credentials credentials.zip --package-name __change__ --package-version ${command:tcb.getNextPackageVersion} --login-to $DOCKER_REGISTRY $DOCKER_LOGIN $DOCKER_PSSWD --canonicalize docker-compose.prod.yml"
                 }
             },
             "args": [
+                "DOCKER_REGISTRY=${DOCKER_REGISTRY:-registry-1.docker.io}",
```
microhobby: is this right?

microhobby (commented):
@andreriesco please take a look on the comments

assets/tasks/common.json
andreriesco: Yes, that is expected for the GPU configuration, which should be automatically filled base on the default board. This was leading to the problem of it being showed if the user is not using the `-vivante` GPU, which confused the user

andreriesco: Also, it was different from the other places where it is used
image

assets/tasks/common.json
andreriesco: Yes, this one is needed because the DockerHub registry can be empty, nut when using it on the `tcb platform push` the command was done in a way that it cannot. So what this does is attribute the `registry-1.docker.io` value (default value for Dockerhub registry) to the `DOCKER_REGISTRY` variable if it is empty

microhobby: @andreriesco Also I think that this behavior is not implemented in the `tasks.ps1` interpreter. So, would be good to have another solution for this.

andreriesco: It is a bash command, but I moved this check to the `tcb-env-setup.sh` script to make it sure the task works

microhobby:
LGTM

PR #154, opened 2023-12-31
Castello/rc common

microhobby:
Update the Community templates to use the RC containers

Missing yet the Gambas template


microhobby:
Happy 2024! 🥳💃

PR #158, opened 2024-01-12
Release 2.4.0

andreriesco:


microhobby:
Hey @andreriesco could you please provide a sum of the work done here in the MR description? Thanks

Edit: nevermind, I'm checking by commit and make more sense now

assets/tasks/common.json
```diff
@@ -641,6 +640,26 @@
                 "color": "terminal.ansiYellow"
             }
         },
+        {
+            "label": "remove-dangling-images",
```
microhobby, line 644: did you saw some impact in the performance? Because, ok that we leave some dangling behind, for the same hashes this should improve the performance on the next push

andreriesco: As far as I observed this task tends to run quickly, because there is not much to remove when something is changed on the source code which is pretty much the last layers that are small.
On the automated tests I saw a very nice improve of performance actually. As in those tests all the projects created have the same name, removing the older one improved the performance, and also prevented the problem of [no space left on the device]

andreriesco: But if you could test it on your side also would be nice to also make sure this task doesn't bring any side effects.

andreriesco: If there are many dangling images at the first time it can take some seconds, but for the second time and beyond it is fast

microhobby:
LGTM

PR #164, opened 2024-01-25
Release 2.4.0

andreriesco:


aspnetBlazor/.vscode/tasks.json
```diff
@@ -138,20 +138,10 @@
             "label": "deploy-torizon-arm",
             "detail": "",
             "hide": true,
-            "command": "scp",
+            "command": "sleep",
```
microhobby, line 141: maybe is better use an `echo ok` than a `sleep 1`?

microhobby:
LGTM!

aspnetBlazor/.vscode/tasks.json
microhobby: let's leave this one for this cycle, and the next one we need to fix to use the wait-a-bit instead and use an echo here.

PR #183, opened 2024-03-06
scripts: torizonIO: Do not limit the number of packages and fleets

microhobby:
The default limit is 50, but we need to get all packages and fleets to be able to search for the target and fleet by name.


andreriesco:
LGTM

PR #189, opened 2024-03-14
Add the possibility to use scripts in deps.json, for check-deps

andreriesco:
Add the possibility to use scripts in the deps.json file, that are
  checked by the check-deps task. Also, add the dotnet-sdk-* and the
  mono-devel install scripts.
  

scripts/checkDeps.ps1
```diff
@@ -65,9 +71,22 @@ if ($_packagesToInstall.Count -gt 0) {
...
         sudo apt-get update
 
+        # ok docker and docker-compose exist so let's check the packages
+        $_installedScrips = New-Object Collections.Generic.List[string]
+
+        # If it is a script, execute the scrip, else apt-get install the package
         foreach ($item in $_packagesToInstall) {
-            sudo apt-get install -y $item
-            
+            if ($item.EndsWith('.sh')) {
```
microhobby: I would not like to force the template developer to use only shell scripts. So, my proposal for this feature would be add a new property on the `deps.json` the `installScripts`, should look like:

```json
{
    "packages": [
        "openssh-client",
        "sshpass",
        "build-essential",
        "make",
        "gdb",
        "gcc"
    ],
    "installScripts": [
        "./install-mono.sh",
        "/home/user/libs/install-lib.ps1",
        "/root/binaryname"
    ]
}
```

This new property should be an array that accepts absolute or relative to `.conf` paths.
Also, would be good to store the specific install scripts on the `.conf` folder of the specific template.

Let me know what you think.

microhobby (changes requested):

scripts/checkDeps.ps1
andreriesco: About the scrips, I had added just powershell and bash, but I agree it is better to have them as a separated property and support any type of script, as long as it is executed in the normal `./$scriptName` way of executing scripts

andreriesco: About them being stored in each template `.conf` folder, it has the good side of being easier to understand where the script is in this templates repo. But it has the downside of maintenance, the `install-dotnet-sdk-8.sh` for example is used on 6 templates, meaning that if we change anything in this script we will need to remember to change it is all of them.
So about this point I am not so sure what would be the best option.

microhobby: hmm, yeah, you have a point. What about using variables then? so, templates developers should be able to do this:

```json
{
    "packages": [
        "openssh-client",
        "sshpass",
        "build-essential",
        "make",
        "gdb",
        "gcc"
    ],
    "installScripts": [
        "./install-mono.sh",
        "${HOME}/.apollox/scripts/install-lib.ps1",
        "/root/binaryname"
    ]
}
```

The main idea here is to be able to detach the workspace from the `.apollox` folder. I think that using variables we can make easy if in some case the user want to detach from the `.apollox` folder by copying the script and changing the `deps.json`.

microhobby: to start let's support only `${HOME}`, and if needed we add more ...

andreriesco: I updated here the commit. I made it in a way that it supports both scripts present just in the template and also common scripts, which are copied from the `scripts/install-deps-scripts` to the `.conf/install-deps-scripts` of the project when the project is created. Also, I added it in a way that of for example the template already have this `.conf/install-deps-scripts/$script` modified to his specific use case (let's say for example that Avalonia uses a modified `install-dotnet-sdk-6.sh` for example) it gives priority to the one already present in the template.

dotnetConsole/.conf/install-deps-scripts/install-dotnet-sdk-8.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+set -e
+
+echo "bbbbb"
```
microhobby: bunnies-what

scripts/checkDeps.ps1
```diff
@@ -40,48 +38,129 @@ if ($? -eq $false || [string]::IsNullOrEmpty($_dockerComposeV)) {
...
     } else {
-        Write-Host -ForegroundColor DarkGreen "👍 $package installed"
+        Write-Host -ForegroundColor DarkGreen "👍 $script dependency installation script executed before for this project on this machine"
     }
 }
 
-# ask if the user want to install the packages that are not installed
-if ($_packagesToInstall.Count -gt 0) {
+# Check if there are any packages to be installed or scripts to be executed
+if ($_packagesToInstall.Count -eq 0 -and $_scriptsToInstall.Count -eq 0) {
```
microhobby: optional, to be better to read:

```powershell
if (
    ($_packagesToInstall.Count -eq 0) -and
    ($_scriptsToInstall.Count -eq 0)
) {
```

scripts/checkDeps.ps1
```diff
@@ -40,48 +38,129 @@ if ($? -eq $false || [string]::IsNullOrEmpty($_dockerComposeV)) {
...
+    Write-Host -ForegroundColor DarkGreen "✅ All packages already installed and installation script executed before for this project on this machine"
```
microhobby: I'm not a big fan of these large messages, could you try to keep it short?

scripts/checkDeps.ps1
```diff
@@ -40,48 +38,129 @@ if ($? -eq $false || [string]::IsNullOrEmpty($_dockerComposeV)) {
...
+
+    # we need to ran the check deps only if it's not ran yet
+    New-Item -Path .conf/ -Name .depok -ItemType File 2>&1 | out-null
```
microhobby: This will not overwrite the scripts names added to the file?

scripts/checkDeps.ps1
```diff
@@ -40,48 +38,129 @@ if ($? -eq $false || [string]::IsNullOrEmpty($_dockerComposeV)) {
...
+
+        }
+    }
+    # ask if the user want to execute the installation scripts that have not been executed before for this project on this machine
+    if ($_scriptsToInstall.Count -gt 0) {
+
+        $_scriptsInstalledOk = $false
+
+        $_installConfirm = Read-Host `
+            "Do you want to try to install the missing debian package dependencies? <y/N>"
```
microhobby: I think it would be better if we could ask just once, here we can treat it as "dependencies", scripts and packages, and then ask once and install everything.

scripts/createFromTemplate.ps1
```diff
@@ -146,6 +146,32 @@ Copy-Item "$templateFolder/../scripts/torizonIO.ps1" "$location/.conf"
...
 
+# Check if there are scripts defined in the .conf/deps.json of the template and, if so,
+# copy them to the .conf of the project
+$_deps = Get-Content "$templateFolder/.conf/deps.json" | ConvertFrom-Json
+
+# If there are installation scripts listed on the .conf/deps.json of the template
+if (($_deps.installDepsScripts.Count -gt 0)) {
+    # Create the .conf/install-deps-scripts if it doesn't exist and there are
+    # installation scripts that have the path .conf/install-deps-scripts
+    if (-not (Test-Path -Path "$location/.conf/install-deps-scripts" )){
```
microhobby: Let's maintain consistency and always use camel case if need more than one word in files ors `installDepsScripts`

scripts/projectUpdater.ps1
```diff
@@ -325,6 +358,19 @@ _openMergeWindow `
     $projectFolder/.conf/tmp/.gitignore `
     $projectFolder/.gitignore
 
+# DEPS.JSON:
+_openMergeWindow `
+    $projectFolder/.conf/tmp/deps.json `
+    $projectFolder/.conf/deps.json
```
microhobby, line 384: Do we really want this? IMHO would be better and more secure to only accept updates on the `deps.json` and scripts without user interaction.

microhobby (changes requested):

scripts/projectUpdater.ps1
andreriesco: This one is mainly because of the addition of the scripts. It is very possible that the use has some script that his project depends on and he will add it there. I see this being used on a project that has more than one developer for example. Once the developer shares the project with the other he wants it to install also some dependency or installation script that his project depends on.

scripts/createFromTemplate.ps1
andreriesco: Sure, I will change it here

scripts/checkDeps.ps1
andreriesco: Sounds good, I will change it here

scripts/checkDeps.ps1
andreriesco: Sure

dotnetConsole/.conf/install-deps-scripts/install-dotnet-sdk-8.sh
andreriesco: Ops, residual echo from testing the update script, I will remove it here

scripts/projectUpdater.ps1
microhobby: ok, make sense. Let's leave this way then

scripts/checkDeps.ps1
andreriesco: I tested here and it does not overwrite, but I will add a check if the file exists before running this command just to be safer.

aspnetBlazor/.conf/deps.json
```diff
@@ -1,7 +1,9 @@
 {
     "packages": [
         "openssh-client",
-        "sshpass",
-        "dotnet-sdk-6.0"
+        "sshpass"
+    ],
+    "installDepsScripts": [
+        ".conf/install-deps-scripts/install-dotnet-sdk-8.sh"
```
microhobby: these are still not in camel case

microhobby (changes requested):

microhobby:
LGTM

PR #195, opened 2024-04-18
python3: Fix and improve pip install on host and containers

andreriesco:
Remove --system-break-packages from pip install and add requirements-local.txt to separate the host pip dependencies from the debug container ones (requirements-debug.txt).

### DIFF
--- a/python3Console/.doc/README.md
+++ b/python3Console/.doc/README.md
@@ -15,7 +15,10 @@ The deployable package is then copied into the running debug container using **s
 
 Remote debugging is performed by running **debugpy** inside the running container on the device, in a task named **start-torizon-debug\${architecture}**, and then attaching to this running container, as described at [VSCode Python remote debugging](https://code.visualstudio.com/docs/python/debugging#_remote-script-debugging-with-ssh). For local debugging, the method used is the one described in the [VSCode documentation for Python Debugging](https://code.visualstudio.com/docs/python/debugging) is used. The tasks that perform the debugging are those present in the **launch.json** file.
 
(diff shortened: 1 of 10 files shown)


microhobby:
@brunoaamello adding you as reviewer, since you point out this.

brunoaamello:
Hello @andreriesco, 

The addition of the `requirements-local.txt` file is a bit different than what I did on https://github.com/toradex/torizon-samples/pull/18. There I install the packages on `requirements-local.txt` and `requirements-debug.txt` on the local venv.

I see your implementation as better, as it gives the user more control over the packages installed on each circumstance. the reason I brought this up is so this difference is known.

brunoaamello:
Regarding the removal of the `--break-system-packages` flag, this change is great and welcome.
The effect is that this template will not work locally unless a virtual environment is created.
Have you confirmed if this is the case?

brunoaamello (approved):
I made two comments to make sure everything is understood about these changes.
The changes are appropriate in my opinion.

andreriesco:
> Regarding the removal of the `--break-system-packages` flag, this change is great and welcome. The effect is that this template will not work locally unless a virtual environment is created. Have you confirmed if this is the case?

Actually it does work if no virtual environment is created (maybe on Debian 12/Ubuntu 24 it is not possible anymore to install python packages outside a venv, so the user will be forced to create one).
 The `--break-system-packages` allows a pip installation of a package outside a venv (meaning directly on the host) even if there is already this package installed on the system through `apt`, therefore "breaking" this system packages installed through apt. By removing it, an error happens if the user tries to pip install a package that is already installed by apt.

andreriesco:
About the `requirements-local.txt`, I implemented this way in this commit exactly for what you said of having more over the packages installed on each circumstances

PR #202, opened 2024-05-22
TCB - docker-in-docker build

ghost:
The existing github actions does not function for a docker-in-docker (`dind`) configuration. Due to the build script using directories with `root:root` ownership a rootless self-hosted runner can not complete this workflow.

The configuration below adds `-wd working_directory` argument to the `tcb-env-setup.sh` script to allow absolute path to be passed directly to the `docker run torizon/torizoncore-builder`. Due to a `dind` setup, the `/workdir` needs to be mounted with an absolute path. Github actions provides variables containing the absolute path for the runner directory, however, this is a well documented bug in the runner since 2022:
- [issue2058](https://github.com/actions/runner/issues/2058)
- [issue406](https://github.com/actions/runner/issues/406)
- [issue785](https://github.com/actions/checkout/issues/785)

A workaround for this is to add a build step creating a file with this abs-path to be referenced later`echo "${{ github.workspace }}" >> abs-path` . Any ENV var set in actions will be replaced at run-time by the runner replacing the ABS path with a relative sandboxed path for the runner. This is why a file is written, not an ENV.

The `scripts/tasks.ps1` now has an additional ENV check to detect if it's in a Actions workflow, and set the abs-path accordingly. I did not handle errors for this `abs-path` file not existing.

| :exclamation:        This is only tested for `tcb-build` stage, no subsequent steps -  `tcb-platform-push-ostree`, or `platform-update-fleet`  :exclamation:|
|-----------------------------------------|

REF: email
Signed-off-by: d.kelly@coligo.ai

### DIFF
--- a/scripts/bash/tcb-env-setup.sh
+++ b/scripts/bash/tcb-env-setup.sh
@@ -43,6 +43,7 @@ tcb_env_setup_cleanup () {
     unset under_windows
     unset user_tag
     unset storage
+    unset working_directory
     unset volumes
     unset network
     unset remote_tags
@@ -95,6 +96,11 @@ tcb_env_setup_usage () {
     echo "      It must be an absolute directory or a Docker volume name. If this"
     echo "      flag is not set, the \"storage\" Docker volume will be used."
     echo ""
+    echo "  -wd: select working directory for docker mount to /workdir"
+    echo "      Pass the directory explicitly for docker run torizon/torizoncore-builder"
+    echo "      to mount as /workdir."
+    echo "      It must be an absolute directory. If this is not set, \$(pwd) with be used"
+    echo ""
     echo "  -n: do not enable \"host\" network mode."
     echo "      Under Linux the tool runs in \"host\" network mode by default allowing"
     echo "      it to operate as a server without explicit port publishing. Under"
@@ -133,6 +139,7 @@ fi
 # Parse flags
 volumes=" -v /deploy "
 storage="storage"
+working_directory="working_directory"
 network=" --network=host "
 if [ $under_windows = "1" ]; then
     # Do not use "host" network mode under Windows/WSL
@@ -144,6 +151,7 @@ do
         -a) source=$2;[ "$2" ]||source="empty"; shift; shift;;
         -t) user_tag="$2";[ "$2" ]||user_tag="empty"; shift; shift;;
         -s) storage="$2";[ "$2" ]||storage="empty"; shift; shift;;
+        -wd) working_directory="$2";[ "$2" ]||working_directory=$(pwd); shift; shift;;
         -d) volumes=" "; shift;;
         -n) network=" "; shift;;
         --) shift; break;;
@@ -162,7 +170,7 @@ then
   tcb_env_setup_check_updated $SCRIPT_PATH
 fi
 
-if [[ $source = "empty" ]] || [[ $user_tag = "empty" ]] || [[ $storage = "empty" ]]
+if [[ $source = "empty" ]] || [[ $user_tag = "empty" ]] || [[ $storage = "empty" ]] || [[ $working_directory = "empty" ]]
 then
     tcb_env_setup_usage
     tcb_env_setup_cleanup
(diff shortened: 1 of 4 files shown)


scripts/bash/tcb-env-setup.sh
```diff
@@ -95,6 +96,11 @@ tcb_env_setup_usage () {
     echo "      It must be an absolute directory or a Docker volume name. If this"
     echo "      flag is not set, the \"storage\" Docker volume will be used."
     echo ""
+    echo "  -wd: select working directory for docker mount to /workdir"
+    echo "      Pass the directory explicitly for docker run torizon/torizoncore-builder"
+    echo "      to mount as /workdir."
+    echo "      It must be an absolute directory. If this is not set, \$(pwd) with be used"
```
microhobby, line 102: > \$(pwd) with be used

I think this was a typo `with` should be `will`, right?

tcb/.github/workflows/build-torizoncore.yaml
```diff
@@ -18,6 +18,15 @@ jobs:
...
           cat abs-path
+      - name: Install build tools in container
+        run: | 
+          apt-get update
+          apt-get install -y wget curl libicu-dev
+          curl -fsSL https://get.docker.com -o get-docker.sh 
+          sh get-docker.sh
+                    
+          wget -nv https://github.com/PowerShell/PowerShell/releases/download/v7.4.2/powershell_7.4.2-1.deb_amd64.deb
+          dpkg -i powershell_7.4.2-1.deb_amd64.deb
```
microhobby, line 29: so, maybe this is needed on your end because you host your own runner. But, on Github Actions the default image already has these dependencies.

Would be nice to first identify whether the dependencies are already installed before installing it.

tcb/.github/workflows/build-torizoncore.yaml
```diff
@@ -6,13 +6,18 @@ jobs:
   build-deploy:
     runs-on: ubuntu-latest
     name: Build & Deploy TorizonCore
+    container: ubuntu:23.04
```
microhobby: Is this needed because of the self-hosted runner?

scripts/bash/tcb-env-setup.sh
ghost: yes ..will be used

tcb/.github/workflows/build-torizoncore.yaml
ghost: this is needed because I'm running the build inside a container, on the shared and self-hosted runners. Both will spin up a docker Ubuntu:23.04 container, mount the local filesystem, and run the build commands. It's good for reproducibility. If you used GitLab, running dind is standard.

tcb/.github/workflows/build-torizoncore.yaml
ghost: no, see this post: https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container

microhobby:
@dk-coligo sorry for the delay, we are in the process of a new release, so, I will back to this after the release. We need to add it to our CI/CD integration tests to make sure that this will not break it the already existent experience.

ghost:
@microhobby we've had to make many other changes to the default scripts. For example, the platform-push stage for the tasks.json. In the[ developer documentation](https://developer.toradex.com/torizon/torizon-platform/torizon-updates/signing-and-pushing-torizoncore-packages-to-torizon-ota/#signing-and-pushing-the-image-to-torizon-cloud), there is a dependency on `tcb unpack` then `tcb platform push`. When I add this to the `depends-on:` parameter of the platform-push task in tasks.json, only the unpack command runs and then exits without running the platform-push. So, `unpack` needs explicit running in the build.yaml stage. I'll comment on the file inline separately so you can find it :)

andreriesco:
LGTM. I will rebase it here and include it in the next 2.6 release. Thanks for the contribution

microhobby:
@dk-coligo I reimplemented the stuff reported here, so instead of adding a `-wd` argument we are setting a `HOST_GITHUB_WORKSPACE` env variable to change the default behavior. This way it make easier on our side to maintain it for all the templates not only for TCB one.
Take a look on 5a6849f3b4d7031483174615d77bee8fc16c96a6 and if this does not fit your needs feel free to reopen this ticket.

PR #204, opened 2024-05-23
Change Rust deps installation to the recommended way, through rustup and revert from version 1.6.0 to 1.5.1 on Slint templates

andreriesco:


scripts/installDepsScripts/installRustDeps.sh
```diff
@@ -0,0 +1,27 @@
...
+declare is_rustup_default=$(rustup show | awk '/\(default\)/')
+
+
+# If Rust comes from rustup, update Rust
+if [[ $is_rustup_default ]]; then
+
+    rustup update
+
+else
+    sleep 1
```
tronical: Why sleep here?

scripts/installDepsScripts/installRustDeps.sh
```diff
@@ -0,0 +1,27 @@
...
+    # Remove the Rust installation comming from apt instead of rustup
+    sudo apt-get remove *rust* -y
```
tronical: Why remove these packages?

rustup will put ~/.cargo/bin at the front of your PATH.

tronical (approved):
Slint bits look good to me. Thanks!

scripts/installDepsScripts/installRustDeps.sh
andreriesco: I added this because the output of the previous `declare is_rustup_default=$(rustup show | awk '/\(default\)/')` command, when the command `rustup show` fails as it does not have rustup installed, was messed with the input of `sudo` asking for the password in the next command. So I added this `sleep 1` to make sure the output of the command and the input asking for the password where nicely shown in sequence in the terminal

scripts/installDepsScripts/installRustDeps.sh
andreriesco: To remove the previous rust packages coming from the Ubuntu source installed through `apt`. To ensure that the rust packages used are always the ones coming from `rustup` and avoid the ones coming from Ubuntu being used unexpectedly, leading to unexpected behaviors. But I don't know much about Rust, so this may not make much sense for Rust. So, is it ok to keep the rust packages installed through `apt`?

microhobby: @andreriesco I think is ok not mess up with the apt packages, let's remove this.

rustConsole/.vscode/launch.json
```diff
@@ -12,7 +12,6 @@
             "environment": [],
             "externalConsole": false,
             "MIMode": "gdb",
-            "miDebuggerPath": "/usr/bin/rust-gdb",
```
microhobby, line 15: but should we not add the path for the new one?

andreriesco: Because the path changes when installing Rust with `rustup`, so this was leading to issues, but if this is removed VSCode gets the default path for Rust, and then it works. Also, this parameter is not needed for the Rust template to work properly.

scripts/installDepsScripts/installRustDeps.sh
andreriesco: Ok, I will remove it

PR #222, opened 2024-08-13
Next to Dev

microhobby:


assets/tasks/common.json
```diff
@@ -672,13 +731,157 @@
...
+                "wait-a-bit"
+            ],
+            "problemMatcher": "$msCompile",
+            "icon": {
+                "id": "flame",
+                "color": "terminal.ansiYellow"
+            }
+        },
+        {
+            "label": "pre-cleanup-debug",
```
andreriesco: Tested here on imx6ull and it seems that it is not bringing down the other containers, which is leading to problems
image

microhobby: This is by design, now it only down the service of the workspace application. This is yet another change to make the development flow faster between one deployment and another. What types of problems are you facing?

andreriesco: It is not bringing down the other containers before running my container, slowing down the execution of the container

andreriesco: It would be nice if we could figure out a way to run the previous command (which is called now `containers-cleanup` I think) at least the first time the person starts a debug session, so it starts with all other containers down and also removes the orphan containers

microhobby: Please keep in mind that you are testing different stacks, this is not the field scenario where the user sticks in a workspace with a specific stack. 

Your user case is from a template developer/tester, so if run this makes sense to your scenario add it to the test pipeline.

PR #223, opened 2024-08-14
 checkSingleProjectsConflicts:Allow arg to bypass manual input, also allow execution from the IDE side. Python improve debug performance. Other fixes

andreriesco:


scripts/multiContainer/checkSingleProjectsConflicts.ps1
```diff
@@ -17,6 +17,7 @@
 )]
 param()
```
microhobby: Maybe is better to add a named param?

scripts/projectUpdater.ps1
```diff
@@ -61,6 +61,14 @@ function _openMergeWindow ($_updatedFile, $_currentFile) {
     }
 }
 
+# Don't allow the update of the project if there is no git repository initiated, to avoid losing track of the changes applied in the update
```
microhobby, line 64: I do not know about this one, perhaps instead of forcing we could show a prompt for the user to accept continuing running

scripts/multiContainer/checkSingleProjectsConflicts.ps1
andreriesco: Sure

scripts/projectUpdater.ps1
andreriesco: Sure, fixed

aspnetBlazor/Dockerfile
```diff
@@ -18,7 +18,7 @@ ARG APP_ROOT=
 
 
 # BUILD ------------------------------------------------------------------------
-FROM mcr.microsoft.com/dotnet/sdk:8.0 AS Build
+FROM mcr.microsoft.com/dotnet/sdk:8.0 AS build
```
microhobby, line 21: as you changed here the `Build` to `build` the `COPY` in line number 55 will fail.
Please check if we are referencing the stage name because we will need to update them all.

microhobby:
LGTM! Thanks @andreriesco

PR #225, opened 2024-08-19
tasks: Update icon and presentation of tasks that were missing this. Also remove unused tasks

andreriesco:


assets/tasks/common.json
```diff
@@ -731,78 +738,7 @@
...
-                "wait-a-bit"
-            ],
-            "problemMatcher": "$msCompile",
-            "icon": {
-                "id": "flame",
-                "color": "terminal.ansiYellow"
-            }
-        },
-        {
-            "label": "pre-cleanup-debug",
```
microhobby, line 778: these should not be removed I think ...

andreriesco: It is actually still there down below, just removed because it was duplicated.

assets/tasks/common.json
```diff
@@ -285,38 +260,6 @@
                 "group": "cloud-related"
             }
         },
-        {
-            "label": "docker-login",
```
microhobby, line 289: The same here.

assets/tasks/common.json
```diff
@@ -132,31 +132,6 @@
                 "group": "cloud-related"
             }
         },
-        {
-            "label": "set-platform-fleet-name",
-            "detail": "",
```
microhobby, line 137: this actually is a util task, it was set a hide true wrongly, maybe because was not description field.

assets/tasks/common.json
microhobby: @andreriesco could you instead remove add a description and set the hide: false ?

andreriesco: Sure

assets/tasks/common.json
andreriesco: Sure, I will fix it

andreriesco: This is being covered by the fill-pipeline-settings task now I believe

PR #227, opened 2024-08-21
templates.json: Remove deprecated dotnetUno and dotnetUnoFrameBuffer.  Also, little fix on createFromTemplate.ps1

andreriesco:
Remove this deprecated templates from templates.json, so they don't showup as options on project creation.


microhobby:
LGTM!

PR #232, opened 2024-08-23
pre-cleanup: Have just one pre-cleanup, that stops debug and release

andreriesco:
Have just one pre-cleanup, that stops debug and release containers, to avoid having the debug container and the release container running at the same time.


microhobby:
LGTM

PR #246, opened 2024-08-29
Fix warning on startup about unrecognized renderer

tronical:
> slint linuxkms backend: unrecognized renderer skia, falling back default

Enable the skia renderer to fix this.


andreriesco:
LGTM

PR #276, opened 2024-10-24
Drop commontorizon from Toradex supported

microhobby:
From feedback is not good to have experimental stuff or release Toradex official stuff “crossover-ed” with experimental stuff. So, we need to drop everything experimental from the Toradex supported templates.

@andreriesco this will be added only on the TOS 6, so this will include also what is not in sync yet from `dev`. This should be the last release on `bookworm` and will enter in maintenance mode.


cppConsole/.vscode/launch.json
```diff
@@ -26,50 +26,6 @@
...
-                {
-                    "description":  "Set Disassembly Flavor to Intel",
-                    "text": "-gdb-set disassembly-flavor intel",
-                    "ignoreFailures": true
-                }
-            ],
-            "preLaunchTask": "deploy-torizon-riscv64"
-        },
         {
             "name": "Torizon AMD64",
```
andreriesco: Also this one can be removed

microhobby: Done in 3763f80171065167be1e385a8aa90dadc92b88f4

microhobby:
@andreriesco please take a look, the tests are ok:
[report.zip](https://github.com/user-attachments/files/17732606/report.zip)

cppConsole/.vscode/launch.json
andreriesco: Just for `cConsole` on this commit, for `cppConsole` it still needs to be removed

README.md
```diff
@@ -6,17 +6,17 @@ This repository maintains the templates used in conjunction with the [VS Code To
 
 > ℹ️ **Toradex Supported Templates** are templates with development stacks commonly used by customers of the Torizon platform and are maintained by the Torizon team at Toradex.
 
-> ⚠️ Support for **x86-64** and **riscv64** architectures are experimental.
+> ⚠️ Support for **x86-64** architectures are experimental.
 
 | TEMPLATE | DESCRIPTION | RUNTIME | LANGUAGE | HW ARCH | FOLDER |
 | -------- | ----------- | ------- | -------- | ------- | ------ |
-| ![](https://raw.githubusercontent.com/toradex/vscode-torizon-templates-documentation/refs/heads/main/thumbnails/cppconsole.png?raw=true) | C++ Makefile Console Application | libc | C++ | ![](assets/img/arm32.png?raw=true&id=2) ![](assets/img/arm64.png?raw=true&id=2) ![](assets/img/riscv64.png?raw=true&id=2) ![](assets/img/x8664.png?raw=true&id=2) | [cppConsole](./cppConsole) |
+| ![](https://raw.githubusercontent.com/toradex/vscode-torizon-templates-documentation/refs/heads/main/thumbnails/cppconsole.png?raw=true) | C++ Makefile Console Application | libc | C++ | ![](assets/img/arm32.png?raw=true&id=2) ![](assets/img/arm64.png?raw=true&id=2) ![](assets/img/x8664.png?raw=true&id=2) | [cppConsole](./cppConsole) |
```
andreriesco: Also the `` needs to be removed from the templates where it has been removed.

scripts/installDepsScripts/installDotnetSDK8.sh
```diff
@@ -50,9 +50,10 @@ if [ "$source" != "https://packages.microsoft.com/$repo/$repo_version/prod" ]; t
     rm packages-microsoft-prod.deb
 
     # Remove the dotnet-sdk installation that doesn't come from the Microsoft source
+    set +e
```
andreriesco, line 53: Does it makes sense to add also this fix on `installDotnetSDK7` script, which is used by Avalonia templates?

cppConsole/Dockerfile
```diff
@@ -12,7 +12,7 @@ ARG APP_ROOT=
 
 
 # BUILD ------------------------------------------------------------------------
-FROM commontorizon/debian-cross-toolchain-${IMAGE_ARCH}:${CROSS_SDK_BASE_TAG} AS build
+FROM torizon/debian-cross-toolchain-${IMAGE_ARCH}:${CROSS_SDK_BASE_TAG} AS build
```
andreriesco, line 15: Here we should use `commontorizon` as an exception, because the container for arm is `debian-cross-toolchain-armhf` instead of `debian-cross-toolchain-arm`. Also, the tags are different for `debian-cross-toolchain-armhf` and `debian-cross-toolchain-arm64`. `3.2.0` for example is not available for `debian-cross-toolchain-armhf, `3.3.0` and beyond is not available for `debian-cross-toolchain-arm64`, one option for this problem would be to use just the major `3` though.

microhobby: This should be handled by 86d22b3

cppConsole/.vscode/launch.json
microhobby: Check it again, there is no `deploy-torizon-riscv64` for `cppConsole`

README.md
microhobby: This should be fixed on 7258c62

scripts/installDepsScripts/installDotnetSDK8.sh
microhobby: yes make sense, added here 3313e68

cppConsole/Dockerfile
microhobby: Some adjustment on build steps need to be also added: 6018428

cppConsole/.vscode/launch.json
andreriesco: Not `riscv64`, the `Torizon AMD64` one

cppConsole/Dockerfile
andreriesco: Sounds good this adjustments

cppConsole/.vscode/launch.json
microhobby: ok, so, `risc-v` was removed from everywhere. `x86_64` only from Toradex supported.

PR #268, opened 2024-10-14
Draft: Torizon 7 Stable release

andreriesco:

PR #278, opened 2024-11-05
CMake C++ GTK 3 Community Template

leograba:
Create a CMake C++ GTK 3 community template (sponsored by Toradex, but not regularly maintained like the Toradex-supported ones).

This template is based on the cmakeConsole and modified inspired on the cppQML for enabling graphics.

This PR is a fork of https://github.com/torizon/vscode-torizon-templates/pull/268 and depends on https://github.com/toradex/vscode-torizon-templates-documentation/pull/16


leograba:
Tested on `Local AMD64` and `Torizon ARMv8`.

This PR is ready for review.

cmakeGTK3/.conf/update.json
```diff
@@ -0,0 +1,14 @@
...
+    {
+        "source": "CMakeLists.txt",
+        "target": "CMakeLists.txt"
+    },
+    {
+        "source": ".vscode/c_cpp_properties.json",
+        "target": ".vscode/c_cpp_properties.json"
+    },
+    {
+        "source": "Dockerfile.sdk",
```
andreriesco: No need for Dockerfile.sdk, it does this check in the projectUpdate.ps1 script. I forgot to remove it from cmakeConsole template. Just did it now

leograba: Got it. It's removed now. Please check.

cmakeGTK3/Dockerfile
```diff
@@ -0,0 +1,110 @@
...
+        ; fi \
+    && \
+    apt-get clean && apt-get autoremove && \
+    rm -rf /var/lib/apt/lists/*
+
+# __deps__
+RUN apt-get -q -y update && \
+    apt-get -q -y install \
+    cmake \
+    libgtk-3-dev:${IMAGE_ARCH} \
```
andreriesco: I just found out that this does not work because for arm32, `IMAGE_ARCH` is `arm`, but `apt` uses `:armhf`. But this is a use case we need to support, not just for this template but for any template. Here there are two possible options:
- Use `if` for the install, like in [here](https://github.com/torizon/vscode-torizon-templates/pull/278/files#diff-9017da37facb87804a1f77573c4cc5998df6f13863ae2077ba807a29b16d8d3fR55) 
- Modify the templates to have a `torizonPackages.json`, instead of creating it on project creation, to have at least the dependencies that depend on the architecture there
Probably will be nice to quickly discuss this on a meeting (like on the grooming)

leograba: > Modify the templates to have a torizonPackages.json, instead of creating it on project creation, to have at least the dependencies that depend on the architecture there

I like it. In fact, in my tests I was using the `torizonPackages.json` but then in the template I didn't use it because I thought this was a file that was mandatory to be left empty in the template. More or less like:

- Keep the base required files for the template to work on the Dockerfile directly
- Keep the `torizonPackages.json` only for the user

But not sure what makes most sense here.

leograba: This should be fixed with https://github.com/torizon/vscode-torizon-templates/pull/278/commits/ddf6eae0bb68d34c3272e4cf200ec9c3a48b8768

cmakeGTK3/Dockerfile
```diff
@@ -50,44 +51,24 @@ WORKDIR ${APP_ROOT}
...
-        ; fi \
-    && \
-    apt-get clean && apt-get autoremove && \
-    rm -rf /var/lib/apt/lists/*
-
-RUN apt-get -y update && \
-    apt-get install -y --no-install-recommends \
-    libgtk-3-0 \
+RUN apt-get -y update && apt-get install -y --no-install-recommends \
+        libgtk-4-1 \
```
andreriesco: Should it be GTK 4 and not GTK 3 here?

andreriesco:
@leograba Please add a gtk3.png image at [assets/img](https://github.com/torizon/vscode-torizon-templates/tree/dev/assets/img)

leograba:
> @leograba Please add a gtk3.png image at [assets/img](https://github.com/torizon/vscode-torizon-templates/tree/dev/assets/img)

This is done

cmakeGTK3/Dockerfile
leograba: Oops, my mistake!

Fixed by https://github.com/torizon/vscode-torizon-templates/pull/278/commits/7ad4338799e9da29f21c0dffd554683a87c8b05f

microhobby:
LGTM thanks @leograba !

PR #295, opened 2025-01-29
Fixes on tasks.py, project-updaters and Github actions pipelines

andreriesco:
If the same env var appeared in a task and in a dependsOn task of this task, like the VSCODE_CMD on tcb-platform-push-ostree, with the tcb-unpack as dependsOn, the env var was not getting updated due to it's order on the recursive function. This commit fixes it.

### DIFF
--- a/scripts/bash/setup-xonsh.sh
+++ b/scripts/bash/setup-xonsh.sh
@@ -13,6 +13,7 @@ fi
 set -e
 
 pipx install xonsh
+pipx ensurepath
 pipx inject xonsh distro
 pipx inject xonsh shtab
 pipx inject xonsh pyyaml
@@ -29,6 +30,9 @@ if ! grep -q "export PATH=\$PATH:\$HOME/.local/bin" ~/.bashrc; then
 fi
 
 # also for .xonshrc itself
+if [ ! -f ~/.xonshrc ]; then
+    touch ~/.xonshrc
+fi
 if ! grep -q "\$HOME/.local/bin" ~/.xonshrc; then
     echo "\$PATH.insert(0, '$HOME/.local/bin')" >> ~/.xonshrc
 fi
--- a/scripts/project-updater.xsh
+++ b/scripts/project-updater.xsh
@@ -151,7 +151,7 @@ if not second_run:
     _git_status: CommandPipeline = {}
     _git_status = !(git -C @(project_folder) status)
 
-    if "fatal: not a git repository" in _git_status.err:
+    if _git_status.err != None and "fatal: not a git repository" in _git_status.err:
         print(
             "❌ fatal: this workspace is not a git repository.",
             color=Color.RED
--- a/scripts/project-updater.xsh
+++ b/scripts/project-updater.xsh
@@ -778,4 +778,10 @@ print("✅ specific files OK", color=Color.GREEN)
 # clean up tmp
 rm -rf @(f"{project_folder}/.conf/tmp")
 
+# update metadata.json
+_project_metadata_file = open(f"{project_folder}/.conf/metadata.json", "w")
+_project_metadata["torizonOSMajor"] = _templates_metadata["TorizonOSMajor"]
+_project_metadata_file.write(json.dumps(_project_metadata, indent=4))
+_project_metadata_file.close()
+
 print("\n✅ Update done", color=Color.GREEN)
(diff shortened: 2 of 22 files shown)


scripts/project-updater.xsh
```diff
@@ -151,7 +151,7 @@ if not second_run:
     _git_status: CommandPipeline = {}
     _git_status = !(git -C @(project_folder) status)
 
-    if "fatal: not a git repository" in _git_status.err:
+    if _git_status.err != None and "fatal: not a git repository" in _git_status.err:
```
microhobby, line 154: in which case the `_git_status.err` is `None`?

scripts/projectUpdater.ps1
```diff
@@ -1,10 +1,6 @@
 
 param()
 
-# include
-. "$env:HOME/.apollox/scripts/utils/formatJson.ps1"
```
microhobby, line 5: instead check if the file exists, then include then

microhobby (changes requested):

scripts/utils/pip/torizon_templates_utils/tasks.py
```diff
@@ -927,4 +927,3 @@ def run_task(self, label: str) -> None:
         if _ret.returncode != 0:
             print(f"> TASK [{label}] exited with error code [{_ret.returncode}] <", color=Color.RED)
             raise RuntimeError(f"Error running task: {label}")
-
```
microhobby, line 930: I do not get the meaning of this change

scripts/project-updater.xsh
andreriesco: When `git -C @(project_folder) status` is a git repository

scripts/utils/pip/torizon_templates_utils/tasks.py
andreriesco: This was a bug present already in [TOS6](https://github.com/torizon/vscode-torizon-templates/pull/294/commits/597c01804ee0ae7795858a249cbe2493401eb1b6), it's the same fix but on the xonsh script. 
But as some other things need to be solved on the pipeline for me to test it on TOS7, I will remove from this MR and add it back on another MR when I have those things fixed

scripts/projectUpdater.ps1
andreriesco: Those files were removed from this repo on TOS7, so this files will never exist

microhobby: Can occur if the user change the repo, branch, tag. Is safer check if they exists then include.

scripts/utils/pip/torizon_templates_utils/tasks.py
microhobby: ok, this I got it. But I'm meaning of this line in particular 930

microhobby: oh ok, you removed the commit.

scripts/bash/setup-xonsh.sh
```diff
@@ -13,6 +13,7 @@ fi
 set -e
 
 pipx install xonsh
+pipx ensurepath
```
microhobby, line 16: what is this fixing?

microhobby:
LGTM, there's only these two comments to clarify.

scripts/bash/setup-xonsh.sh
andreriesco: This was a suggestion from the `pipx install xonsh`, which suggested adding this to add xonsh to the `PATH` variable:
```
installing xonsh...
⚠️  Note: '/opt/pipx_bin' is not on your PATH environment variable. These apps
    will not be globally accessible until your PATH is updated. Run `pipx
    ensurepath` to automatically add it, or manually modify your PATH in your
    shell's config file (e.g. ~/.bashrc).
 ```

scripts/projectUpdater.ps1
andreriesco: In this cases it may happen, I will add the check

andreriesco: Added

scripts/bash/setup-xonsh.sh
microhobby: this need some kind of password or privilegies?

microhobby: ok, so, I check and this does not need privilegies, so it's ok to have it here.

PR #297, opened 2025-02-06
all: Remove the id_rsa id_rsa.pub

microhobby:
We do not use these anymore for none of the templates, so let's remove them from the repository.


andreriesco:
LGTM

PR #298, opened 2025-02-21
Some fixes and improvements (mainly related to the CI/CD pipelines)

andreriesco:


scripts/apply-ci-settings-file.xsh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env xonsh
+
+# Copyright (c) 2025 Toradex
+# SPDX-License-Identifier: MIT
+
+##
+# This script is used to verify the sanity of the CI environment.
+# Is useful to show to the user the env that should be set
+# and fail fast if something is missing.
```
microhobby: The description here was from the other script.

scripts/apply-ci-settings-file.xsh
```diff
@@ -0,0 +1,52 @@
...
+##
+
+# use the xonsh environment to update the OS environment
+$UPDATE_OS_ENVIRON = True
+# Get the full log of error
+$XONSH_SHOW_TRACEBACK = True
+# this script should handle the subprocess errors
+$RAISE_SUBPROC_ERROR = False
```
microhobby: Was this done intentionally? The script should not fail if some of the commands fail?

microhobby (changes requested):

assets/github/workflows/build-application.yaml
```diff
@@ -18,7 +18,7 @@ jobs:
           cat abs-path
 
       - name: Initial Setup
-        shell: xonsh {0}
+        shell: bash
```
microhobby, line 16: I think that we need to apply these changes to the gitlab also. Please take a look.

assets/tasks/common.json
```diff
@@ -145,7 +145,7 @@
                     "TCB_CLIENTSECRET": "${command:tcb.clientSecret}",
                     "TCB_PACKAGE": "__change__",
                     "TCB_FLEET": "${command:tcb.fleetName}",
-                    "TORIZON_ARCH": "${command:torizon_arch}"
+                    "TORIZON_ARCH": "${input:archList}"
```
microhobby: Here we should maintain the `command:torizon_arch` as this will show the options and save it on settings.json

microhobby (changes requested):

scripts/apply-ci-settings-file.xsh
andreriesco: True, will update it.

scripts/apply-ci-settings-file.xsh
andreriesco: No, it should fail, I didn't realize it was wrong, will fix it

assets/github/workflows/build-application.yaml
andreriesco: Sure, will take a look

assets/tasks/common.json
andreriesco: Makes sense, will update it

assets/github/workflows/build-application.yaml
andreriesco: Just realized there is no straight forward way to do this in `.gitlab-ci.yml`. But [bash is already the default shell on Gitlab CI](https://docs.gitlab.com/runner/shells/)

assets/gitlab/.gitlab-ci.yml
```diff
@@ -82,14 +60,13 @@ push-package-to-platform:
...
     TASKS_OVERRIDE_ENV: "False"
 
   script: |
     echo $PLATFORM_CREDENTIALS | base64 -d > credentials.zip
     echo '
 
-      xonsh ./.vscode/tasks.xsh run run-torizon-binfmt
       xonsh ./.vscode/tasks.xsh run tcb-platform-publish
 
     ' | xonsh-wrapper.xsh
```
microhobby: I though that the idea was to use bash directly, no? So, I believe that here we do not need the `xonsh-wrapper` anymore.

microhobby (changes requested):

scripts/create-docker-compose-production.xsh
```diff
@@ -171,7 +170,7 @@ print("✅ Image rebuild and tagged", color=Color.GREEN)
 # push it
 print(f"Pushing {os.environ['DOCKER_LOGIN']}/{_image_name}:{_tag} ...")
 
-echo @(_docker_psswd) | docker login --username @(_docker_login) --password-stdin
+xonsh ./.vscode/tasks.xsh run @(f"docker-login")
```
microhobby: There is little detail here, as this is a string that is not being formatted with other "Python space" variables we could use it directly:

```xonsh
xonsh ./.vscode/tasks.xsh run docker-login
```

scripts/check-ci-env.xsh
```diff
@@ -52,9 +55,10 @@ def _goto_error():
 
 _missing_env_var_settings = False
 _missing_env_var_secrets = False
+_missing_env_var_file_path = False
 
 # check if we are running in a GitLab CI or GitHub Actions environment
-if "GITLAB_CI" in os.environ or "CI" in os.environ:
+if "GITLAB_CI" not in os.environ or "CI" in os.environ:
```
microhobby: I do not get it, why this does not work for `GITLAB_CI` but work for `CI`, should not be `not or not` in the case?

assets/gitlab/.gitlab-ci.yml
```diff
@@ -29,6 +29,7 @@ initial-setup:
   script: |
     echo '
 
+      xonsh .conf/validate-json.xsh
```
microhobby: little detail, follow the same pattern for the `./.vscode` so could be nice to use `./.conf`

assets/gitlab/.gitlab-ci.yml
```diff
@@ -10,80 +10,74 @@ variables:
...
+# Step 2: Create and Publish Image
+publish_image:
+  stage: publish
+  image:
+    name: torizonextras/torizon-dev:dev
+    docker:
+      user: root
+    entrypoint: [""]
+  script:
+    - bash -c 'echo $PLATFORM_CREDENTIALS | base64 -d > credentials.zip'
```
microhobby: Is really necessary to use `bash -c` instead of using the command directly?

scripts/utils/pip/torizon_templates_utils/tasks.py
```diff
@@ -918,6 +918,8 @@ def run_task(self, label: str) -> None:
         # execute the task
         _cmd_join = f"{_cmd} {' '.join(_args)}{_is_background}"
 
+        print(f"Command that will be executed: {_cmd_join}")
```
microhobby: this should be only for debug check the line below

microhobby (changes requested):

scripts/utils/pip/torizon_templates_utils/tasks.py
```diff
@@ -600,7 +600,7 @@ def __check_tcb_inputs(self, env: List[str]) -> List[str]:
                         except KeyError:
                             raise RuntimeError("Error replacing variable tcb.outputTEZIFolder, make sure the tcbuild.yaml has the output.easy-installer.local property")
 
-                        value = value.replace("${{command:tcb.outputTEZIFolder}}", _tezi_folder)
+                        value = value.replace(f"${{command:tcb.outputTEZIFolder}}", _tezi_folder)
```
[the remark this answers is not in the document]
microhobby: make sure that the value is being set correctly here please

assets/gitlab/.gitlab-ci.yml
microhobby: Could you please open a new MR to edit the `torizonextras/torizon-dev:dev` to symlink the bash to sh?

andreriesco: Sure

microhobby:
LGTM

PR #300, opened 2025-02-28
Specify the docker registry as a parameter to "docker login".

drewmoseley:
It seems that the registry needs to be on the command-line for non-DockerHub registries.  I verified this using the python template and ghcr.io.

### DIFF
--- a/assets/tasks/common.json
+++ b/assets/tasks/common.json
@@ -285,7 +285,8 @@
                 "login",
                 "--username",
                 "${command:docker_login}",
-                "--password-stdin"
+                "--password-stdin",
+                "${command:docker_registry}"
             ],
             "dependsOrder": "sequence",
             "icon": {


drewmoseley:
@microhobby I'm not sure about the branches and process here. I saw this issue using the released version of the IDE extension so I _think_ the bookworm branch is right.

I also saw that `scripts/createDockerComposeProduction.ps1` seems to have a similar issue but it appears that the fix there is more involved and a proper fix is already underway for dev. I think that is ok for now; I don't know of any users of the generic dockerfile template that are having this particular issue.

andreriesco:
Hello @drewmoseley ,
Coincidentally I have also been working on the `docker-login` task on [this PR](https://github.com/torizon/vscode-torizon-templates/pull/298/files#diff-8abebfd7c4c9d2b21abadb5f73d2b6d4b4d6c7b5af099614cde8c05dafdd0cc2R78). So, I will close this PR since this fix will be added by the ohter PR there

drewmoseley:
@andreriesco I was able to test this against both hub.docker.com and ghcr.io with the Torizon 6/release version of the extension.  All is good.

microhobby:
LGTM thanks @drewmoseley for the contribution

PR #306, opened 2025-03-18
Fixes on CI and on tasks.py

andreriesco:


scripts/utils/pip/torizon_templates_utils/tasks.py
```diff
@@ -894,18 +894,19 @@ def run_task(self, label: str) -> None:
         if self.__gitlab_ci:
             _cmd = self.__replace_docker_host(_cmd)
 
-        # inject env
+        # Set env var for the task, adding the envs present on the task
+        task_env = os.environ.copy()
```
microhobby, line 897: this one is a trick one, because ok, it solves this issue when we need to set it by a task, but and if we set the env that is used on this task, it will overwrite it anyway or will use the env that I set before run tasks.xsh?

tcb/.vscode/settings.json
```diff
@@ -16,7 +16,7 @@
   "torizon_app_root": "/home/torizon",
   "docker_tag": "v0.0.0",
   "tcb.packageName": "__change__",
-  "tcb.version": "3.11.0",
+  "tcb.version": "early-access",
```
microhobby: I think we must use the `3.12.0` here, I confirmed with Jeremias that this version works on TOS 7.

microhobby (changes requested):

microhobby:
@andreriesco LGTM, only a small detail on the TCB version and that's it

tcb/.vscode/settings.json
andreriesco: Sure, updated

PR #312, opened 2025-03-27
cppQML: Remove Qt for Python extension recommendation

brunoaamello:
This recommendation is not needed as this is a C++ template.

### DIFF
--- a/cppQML/.vscode/extensions.json
+++ b/cppQML/.vscode/extensions.json
@@ -1,7 +1,6 @@
 {
     "recommendations": [
         "ms-vscode.cpptools",
-        "twxs.cmake",
-        "seanwu.vscode-qt-for-python"
+        "twxs.cmake"
     ]
 }


leograba:
As far as I remember, this extension has some nice QML-related features even for the C++ template.

brunoaamello:
> As far as I remember, this extension has some nice QML-related features even for the C++ template.

Ah, that explains why this was here in the first place.

Looking into this, the nice feature is likely the QML preview.
However, it does not work unless `pyside6-qml` is present in the current python environment.
Such environment is not setup by this template as it is not a Python template.

I don't know if `pyside6-qml` is available by default from the package manager from our target distributions.
If it is I think this PR can be safely closed.
If not, I don't see what adding this recommendation brings to users of this template.

In Debian 12 it is not, but I understand that it is not one of the target distributions.

leograba:
This is nice to know. It's beyond my knowledge from now on, though, so let's see what Andre and Castello have to say.

microhobby:
Hey @brunoaamello thanks for looking at this. Yes, the extension was added to provide syntax highlighting for the QML documents. But as you mentioned, this works a bit weird if you do not have the pyside-6 on your system, or python env. I will accept you MR, but for your knowledge, we should work together with Qt to integrate the new extension suite from Qt: https://marketplace.visualstudio.com/items?itemName=TheQtCompany.qt soon.

PR #314, opened 2025-03-28
python3Console: Fix Dockerfile print on output

andreriesco:
On python the output is not printed on the Docker output while the code is being executed if the unbuffered option is not added when executing the file:
https://github.com/moby/moby/issues/12447#issuecomment-93932924


microhobby:
LGTM

PR #331, opened 2025-06-11
scripts: use tcp://docker:2375 only on GitLab CI with Docker executor

lucas-p-bernardes:
GitLab CI sets the GITLAB_CI environment variable even when using the 'shell' executor. In such cases, the tcp://docker:2375 endpoint is not available, causing the scripts to fail in apollox tests due to 'Docker is not running! Error cause: Not configured'.

Fix: In addition to checking for GITLAB_CI, script now also check for the existence of /.dockerenv, which is only present inside Docker containers. This ensures DOCKER_HOST is only set when running under the Docker executor, avoiding failures in shell-based jobs.

Related-to: TIE-527

### DIFF
--- a/scripts/check-deps.xsh
+++ b/scripts/check-deps.xsh
@@ -27,8 +27,8 @@ from torizon_templates_utils.colors import Color,BgColor,print
 # clean the workspace set device default to use the local docker engine
 $DOCKER_HOST = ""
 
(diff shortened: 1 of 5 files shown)


microhobby:
LGTM

microhobby:
Thanks @lucasbernardestoradex !

PR #345, opened 2025-07-21
tasks: Drop usage of the registry to host -> remote exchange

microhobby:
Instead let's use a local store folder to exchange the container image with the remote host. Using the docker save/load commands combined with rsync to perform optimal transfer of the image.

Another gain point is the ability to not use compressed layers, which can be useful when using cloud registries, but for local development it is not needed and can save some processing time under our more less powerful machines.


LeoBelmont:
Good code cleanup overall, LGTM.

assets/tasks/common.json
```diff
@@ -800,6 +800,58 @@
...
+            "hide": true,
+            "command": "DOCKER_HOST=",
+            "type": "shell",
+            "options": {
+                "cwd": "${workspaceFolder}/.store-__container__"
+            },
+            "args": [
+                "docker",
+                "save",
+                "localhost:5002/__container__-debug:${config:torizon_arch}",
```
MiKom, line 830: it probably shouldn't have the `-debug` part in the container name. The build exports it without `-debug`

MiKom: See #350

PR #359, opened 2025-07-30
vscode: ci-tests: Add pylint-xonsh linter step to our CI

guilhermelsousatoradex:
Related-to: TIE-1029


microhobby:
ok, the addition of the linter was successful, but as you see now there is a lot of issues on the code to be fixed or adjusted

microhobby:
Thanks for the work @guilhermelsousatoradex I'm droping this for now, but I will revisit this when we fix the issues on the linter side.

PR #364, opened 2025-08-19
common: Fix "containers-cleanup" task being hidden

andreriesco:


microhobby:
So, the issue here is not that this need to be removed but that the `hide` was set to false wrongly. Please set the `hide` to false

andreriesco:
Sure, will fix it

PR #371, opened 2025-08-27
tcb: Fix command and setting, and add tcb-build-data-online task

andreriesco:


tcb/.vscode/tasks.json
```diff
@@ -376,7 +415,7 @@
             "options": {
                 "cwd": "${workspaceFolder}",
                 "env": {
-                    "VSCODE_CMD": "--verbose platform provisioning-data --credentials credentials.zip --online-data ${command:tcb.onlineData}"
+                    "VSCODE_CMD": "--verbose platform provisioning-data --credentials credentials.zip --online-data DEFAULT"
```
microhobby, line 418: should we not use the tdb.onlineData here too?

andreriesco: Initially I thought yes, but I talked to Lucas and he told me that is this command the value should always be `DEFAULT` . Also, when I try to change it, it gives me this error
```
2025-08-27 14:15:13,242 - torizon.tcbuilder.cli.platform - ERROR - Error: Currently the only supported client-name is "DEFAULT" (aborting).
```

PR #372, opened 2025-09-05
Update DOCKER_PSSWD to DOCKER_PASSWORD and QT_LICENSE_PSSWD to QT_LICENSE_PASSWORD

andreriesco:
Update DOCKER_PSSWD to DOCKER_PASSWORD and QT_LICENSE_PSSWD to QT_LICENSE_PASSWORD, to match docker_password and qt_license_password names
Related-to: TIE-1103


scripts/create-docker-compose-production.xsh
```diff
@@ -247,3 +247,5 @@ yaml.dump(
 _f_ref.close()
 
 print("✅ docker-compose.prod.yml generated", color=Color.GREEN)
+
+xonsh ./.vscode/tasks.xsh run template-specific-final-task
```
microhobby: I'd rather call the initial here at the script too, because have only the initial there on the json can make things confusing.

microhobby (changes requested):

scripts/create-docker-compose-production.xsh
andreriesco: I tried using here at the script, but then I ran into the problem of the script not solving the secrets set on VSCode settings.

microhobby: so, it means that the `tasks.xsh` or the caller of the task `create-docker-compose-production` is not passing it for the `tasks.json` take a look

andreriesco: By default, `tasks.xsh` should be solving VSCode secrets set on `settings.json`?

microhobby: no, the secrets should have been set as an environment variable

microhobby (changes requested):
I'd rather call the initial here at the script too, because have only the initial there on the json can make things confusing.

andreriesco:
> I'd rather call the initial here at the script too, because have only the initial there on the json can make things confusing.

But by calling it in the script makes it not possible to get the VSCode secret, leading to the same problem

scripts/env-vars-json.xsh
```diff
@@ -0,0 +1,127 @@
+#!/usr/bin/env python3
+"""
+Script to save environment variables to a JSON file.
+Takes environment variable names as command line arguments and saves their values to JSON.
```
microhobby: I do not like this approach. Seems complicated to maintain and adds a higher risk of leaking the secrets.

I think this should be done on the extension side instead.

PR #376, opened 2025-09-23
Little fixes

andreriesco:
Related-to: TIE-967

### DIFF
--- a/aspnetBlazor/.gitignore
+++ b/aspnetBlazor/.gitignore
@@ -6,3 +6,4 @@ credentials.zip
 
 .conf/.depok
 .conf/.docok
+.store-*
--- a/assets/tasks/common.json
+++ b/assets/tasks/common.json
@@ -845,7 +845,11 @@
                 ".store-__container__/",
                 "${config:torizon_login}@${config:torizon_ip}:/home/${config:torizon_login}/.store-__container__"
             ],
-            "dependsOrder": "sequence"
+            "dependsOrder": "sequence",
+            "presentation": {
+                "focus": true,
+                "group": "build-execution"
+            }
         },
         {
             "label": "build-container-torizon-debug-arm64",
--- a/assets/tasks/common.json
+++ b/assets/tasks/common.json
@@ -1249,7 +1249,7 @@
                 "-o",
                 "StrictHostKeyChecking=no",
                 "${config:torizon_login}@${config:torizon_ip}",
-                "cd /home/${config:torizon_login} && tar cf - -C .store-__container__/ . | docker load"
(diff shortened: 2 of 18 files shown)


aspnetBlazor/.gitignore
```diff
@@ -5,4 +5,5 @@ credentials.zip
 *.lock.yml
 
 .conf/.depok
-.conf/.docok
\ No newline at end of file
+.conf/.docok
+.store-__container__
```
microhobby: make it with a wildcard

microhobby (changes requested):

dotnetAvaloniaFrameBuffer/.vscode/settings.json
```diff
@@ -1,5 +1,5 @@
 {
-    "torizon_debug_ssh_port": "2224",
+    "torizon_debug_ssh_port": "2223",
```
microhobby: why this change? on the `templates.json`, it says 2224

assets/tasks/common.json
```diff
@@ -845,7 +845,14 @@
                 ".store-__container__/",
                 "${config:torizon_login}@${config:torizon_ip}:/home/${config:torizon_login}/.store-__container__"
             ],
-            "dependsOrder": "sequence"
+            "dependsOrder": "sequence",
+            "presentation": {
+                "echo": true,
+                "reveal": "always",
+                "focus": true,
+                "panel": "shared",
```
microhobby: `echo` `reveal` and `panel` are using the default values, please remove them from here

microhobby (changes requested):

aspnetBlazor/.gitignore
andreriesco: OK

dotnetAvaloniaFrameBuffer/.vscode/settings.json
andreriesco: It was failing with port 2224 for me. Than I changed to 2223 (the same as dotnetAvalonia) and it worked

assets/tasks/common.json
andreriesco: Sure

dotnetAvaloniaFrameBuffer/.vscode/settings.json
microhobby: I had tested and normal behavior, and the 2223 is already in use for other template. So, please drop this commit.

aspnetBlazor/.gitignore
microhobby: no no, like `.store-*` or something like this

PR #389, opened 2025-10-06
tasks: common: Back to registry but with tunneling

microhobby:
The store mechanism was not working was expected with bigger layers.
So, we need to go back to the registry mechanism, but now with
tunneling, so we can work around possible network issues and firewalls
between Windows and the target.

Signed-off-by: Matheus Castello <matheus@castello.eng.br>


microhobby:
LGTC

PR #395, opened 2025-10-15
common.json: Fix wait-a-bit-pull form arm32

andreriesco:


assets/tasks/common.json
```diff
@@ -1294,6 +1294,37 @@
                 "group": "build-execution"
             }
         },
+        {
+            "label": "wait-a-bit-pull",
```
microhobby, line 1298: if this is being annoying I think that we need to first check the service instead of add a dumb sleep.

microhobby (changes requested):

PR #396, opened 2025-10-15
tasks: common: Add tunnel-check task

microhobby:
We are adding weird sleeping loops to wait for the registry to be ready, but a proper check is better.

### DIFF
--- a/scripts/tunnel-check.xsh
+++ b/scripts/tunnel-check.xsh
@@ -25,10 +25,12 @@ from torizon_templates_utils.errors import Error,Error_Out
 
 
 
-if len(sys.argv) < 4:
-    print("Something went wrong, not enough arguments")
-    print("Report on https://github.com/torizon/vscode-torizon-templates/issues")
-    Error_Out("", Error.EINVAL)
+if len(sys.argv) == 4:
+    Error_Out(
+        f"Error: Expected 4 arguments, but got {len(sys.argv)}.\n" +
+        "Report on https://github.com/torizon/vscode-torizon-templates/issues",
+        Error.EINVAL
+    )
 
 
 torizon_psswd = get_arg_not_empty(1)
@@ -36,12 +38,12 @@ torizon_ssh_port = get_arg_not_empty(2)
 torizon_user = get_arg_not_empty(3)
 torizon_ip = get_arg_not_empty(4)
 
-max_attempts = 15
-timeout_seconds = 5
-sleep_interval = 1
+MAX_ATTEMPTS = 15
+TIMEOUT_SECONDS = 5
+SLEEP_INTERVAL = 1
 
 
-for i in range(1, max_attempts + 1):
+for i in range(1, MAX_ATTEMPTS + 1):
     try:
         # Use curl to check registry availability
         # --silent: suppress progress output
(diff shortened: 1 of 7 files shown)


scripts/tunnel-check.xsh
```diff
@@ -0,0 +1,78 @@
...
+
+
+torizon_psswd = get_arg_not_empty(1)
+torizon_ssh_port = get_arg_not_empty(2)
+torizon_user = get_arg_not_empty(3)
+torizon_ip = get_arg_not_empty(4)
+
+max_attempts = 15
+timeout_seconds = 5
+sleep_interval = 1
```
LeoBelmont: This is a bit of a nitpick but since these are constants it would be nice if they were all uppercase to follow [PEP8](https://peps.python.org/pep-0008/#constants)

scripts/tunnel-check.xsh
```diff
@@ -0,0 +1,78 @@
...
+                    @(torizon_user)@@(torizon_ip) curl --silent --max-time @(timeout_seconds) http://localhost:5002/v2/_catalog \
+            )
+
+        if result.returncode == 0:
+            print('Registry ready')
+            sys.exit(0)
+
+    except Exception as e:
+        print(f"Exception occurred: {e}")
+        pass
```
LeoBelmont: This "pass" could be removed as it is redundant in this context

scripts/tunnel-check.xsh
```diff
@@ -0,0 +1,78 @@
...
+import json
+import time
+import subprocess
+from torizon_templates_utils.args import get_arg_not_empty,get_optional_arg
+from torizon_templates_utils.errors import Error,Error_Out
+
+
+
+if len(sys.argv) < 4:
+    print("Something went wrong, not enough arguments")
```
LeoBelmont: This error log can be a bit misleading since it will only fail when the amount of args isn't what we expect. Maybe we could rephrase it to something like f"Error: Expected x arguments, but got {len(sys.argv)}."

scripts/tunnel-check.xsh
```diff
@@ -0,0 +1,78 @@
...
+import sys
+import json
+import time
+import subprocess
+from torizon_templates_utils.args import get_arg_not_empty,get_optional_arg
+from torizon_templates_utils.errors import Error,Error_Out
+
+
+
+if len(sys.argv) < 4:
```
LeoBelmont: We expect a specific amount of arguments here, so would it be better to use "equal" instead of "less" than 4?

LeoBelmont (changes requested):

scripts/tunnel-check.xsh
```diff
@@ -0,0 +1,78 @@
...
+
+    except Exception as e:
+        print(f"Exception occurred: {e}")
+        pass
+
+    print(f"Attempt {i}/{max_attempts}: waiting for registry...")
+    time.sleep(sleep_interval)
+
+
+
```
LeoBelmont, line 73: Another nitpick but could delete a blank line here

LeoBelmont (changes requested):

LeoBelmont (approved):
LGTM

PR #397, opened 2025-10-16
create-docker-compose-prod: Add template-specific-<initial|final>-tasks

andreriesco:


scripts/create-docker-compose-production.xsh
```diff
@@ -256,3 +258,9 @@ yaml.dump(
 _f_ref.close()
 
 print("✅ docker-compose.prod.yml generated", color=Color.GREEN)
+
+# run template-specific-final-task
```
microhobby, line 262: I'm seeing the `$TASKS_OVERRIDE_ENV = False`
after this call, will this mess up with the `template-specific-final-task` call?

andreriesco: I believe so. I will update it for the template-specific-final-task

microhobby (approved):
LGTM

PR #399, opened 2025-10-17
Fixes on project-updater and on wait_sync setting on multiroot wait_sync

andreriesco:
Related-to: TIE-1130
Related-to: TIE-1132
Related-to: TIE-1136
Related-to: TIE-1145

### DIFF
--- a/scripts/project-updater.xsh
+++ b/scripts/project-updater.xsh
@@ -727,10 +727,11 @@ if not os.path.exists(f"{project_folder}/.doc"):
     mkdir -p @(f"{project_folder}/.doc")
 
 for file in os.listdir(f"{project_folder}/.conf/tmp/.doc"):
-    _open_merge_window(
-        f"{project_folder}/.conf/tmp/.doc/{file}",
-        f"{project_folder}/.doc/{file}"
-    )
+    cp -rf \
+    @(f"{project_folder}/.conf/tmp/.doc/{file}") \
+    @(f"{project_folder}/.doc/{file}")
+
+print("✅ .doc files", color=Color.GREEN)
 
 # TASKS.JSON
 _open_merge_window(


scripts/project-updater.xsh
```diff
@@ -140,6 +140,19 @@ def _open_merge_window(to_update, current):
...
+            with open(to_update, "r") as f:
+                content = f.read()
+
+        except (UnicodeDecodeError):
+            print(f"File {to_update} is not readable and cannot be opened on diff tool",
+            color=Color.YELLOW
+            )
+            _iam_sure = input("Do you want to update it by replacing it? [y/n] ")
+            if _iam_sure == "y":
+                cp -f @(to_update) @(current)
```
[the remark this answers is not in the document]
andreriesco: Actually this is intentional. If the file is not readable, it will still show the diff tool but without showing the content instead of failing. And if the file is an image for example, it will show the comparison between images, so it makes sense to present the diff tool first

aspnetBlazor/.doc/README.md
```diff
@@ -21,4 +21,4 @@ The source code of the template is inspired by the one created following the [in
 
 The Blazor is a web application that needs a browser to show the GUI. If you want to show the GUI in a screen connected to the board you can launch the `Torizon Browser ARMv8` debug session. This will run the Blazor application and also a [Cog](https://github.com/Igalia/cog) browser to open the GUI in the board connected screen.
 
-The Docker compose service to run for production that have the Cog browser is `__change__-browser`.
+The Docker compose service to run for production that has the Cog browser is `${project_name}-browser`.
```
microhobby: I do not get it. Why are we changing this? This will have the change from the project creation, so it seems right

scripts/build-multi-root.xsh
```diff
@@ -52,7 +52,9 @@ with open(compose_path, "r") as f:
     compose_yaml = yaml.safe_load(f) or {}
 compose_yaml["services"] = {}
 
-factor = 0
+# Start at -1 because the multiRoot folder itself is not counted as a project
+# as it doesn't have wait_sync setting
+factor = -1
```
LeoBelmont: This could be a problem in the future in case there is another template without wait_sync. Here I'd prefer if we set a "hasWaitSync = false" parameter to the multiRoot templates.json and check it before we add to the variable. We don't need to set this parameter for all templates, if it's not set then we assume true.

LeoBelmont (changes requested):

microhobby (changes requested):

aspnetBlazor/.doc/README.md
andreriesco: Made this because README file was not being correctly updated on `project-updater`. But now reverted this and made a fix that properly updates it on project-updater

scripts/build-multi-root.xsh
andreriesco: Sure, added a better way to fix it, by updating the factor just if it has a wait_sync setting

LeoBelmont (approved):

scripts/project-updater.xsh
```diff
@@ -705,6 +701,15 @@ os.chdir(_old_location)
 # ---------------------------------------------------------------------- .VSCODE
 print("Diff .vscode files ...", color=Color.YELLOW)
 
+if not os.path.exists(f"{project_folder}/.doc"):
+    mkdir -p @(f"{project_folder}/.doc")
+
+for file in os.listdir(f"{project_folder}/.conf/tmp/.doc"):
+    _open_merge_window(
```
microhobby: Hmm, I do not think that we should bother users with a merge for `.doc`, we should simple update it. What do you think @LeoBelmont ?

andreriesco: Sounds good

microhobby (approved):

PR #450, opened 2026-04-09
Fix ssh-check task

MiloszKosobuckiKDAB:
It connects inside the container and inside the container we run as `$config:torizon_run_as` and not as `$config:torizon_login` which is the login to the host OS.

It accidentally worked for others because I'm probably the only soul in the universe that runs my application as a different user than "torizon".


LeoBelmont (approved):
LGTM, lots of thanks for the contribution @MiloszKosobuckiKDAB

PR #456, opened 2026-05-06
Slint: Upgrade to 1.16.1

tronical:
These two commits upgrade the Slint C++ and Rust template to the latest release.


microhobby:
@tronical thanks!

PR #460, opened 2026-05-21
support for arm64 fixes + fixes for latest xonsh

microhobby:
Signed-off-by: Matheus Castello <matheus@castello.eng.br>

### DIFF
--- a/scripts/bash/setup-xonsh.sh
+++ b/scripts/bash/setup-xonsh.sh
@@ -2,6 +2,19 @@
 
 echo "🐚 SETUP XONSH"
 
+_XONSH_TARGET="0.23.7"
+
+function _check_xonsh_update {
+    # make sure we are also updating to the target version of xonsh
+    _XONSH_CURRENT=$(pipx list --short 2>/dev/null | grep "^xonsh " | awk '{print $2}')
+
+    if [ "$_XONSH_CURRENT" = "$_XONSH_TARGET" ]; then
+        echo "xonsh $_XONSH_TARGET already installed, skipping"
+    else
+        pipx install --force-reinstall xonsh==$_XONSH_TARGET
+    fi
+}
+
 function _do_injections {
     pipx inject xonsh distro
     pipx inject xonsh shtab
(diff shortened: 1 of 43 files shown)


scripts/validate-deps-running.xsh
```diff
@@ -125,7 +125,7 @@ if "enabled" not in _cmd_ret_reg.out:
 # FIXME: we check for arm64 and arm32 for now, if a new arch is introduced
 # we need to add the check here
 def _retry_enable_binfmt(_arch: str):
-    docker run --rm -it --privileged torizon/binfmt:latest
+    docker run --rm -it --privileged pergamos/binfmt:9.0.2
```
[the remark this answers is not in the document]
microhobby: Done

scripts/validate-deps-running.xsh
```diff
@@ -125,7 +125,7 @@ if "enabled" not in _cmd_ret_reg.out:
...
```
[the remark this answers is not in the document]
microhobby: But then it infer that `pergamos/binfmt:9.0.2` is not trusted, what's is not true. We need a new binfmt that works on arm64 and x86_64 and the pergamos/binfmt does that

scripts/bash/setup-xonsh.sh
```diff
@@ -2,6 +2,19 @@
...
+function _check_xonsh_update {
+    # make sure we are also updating to the latest version of xonsh
+    _XONSH_CURRENT=$(pipx list --short 2>/dev/null | grep "^xonsh " | awk '{print $2}')
+
+    if [ "$_XONSH_CURRENT" = "$_XONSH_TARGET" ]; then
+        echo "xonsh $_XONSH_TARGET already installed, skipping"
+    else
+        pipx install --force-reinstall xonsh==$_XONSH_TARGET
+    fi
+}
```
LeoBelmont, line 16: Why are locking the xonsh version specifically to 0.23.7? Why not this version and newer?

microhobby: All these changes are caused by newer versions of Xonsh being released, and then new users are facing a lot of warnings on their flow that are polluting it logs. Sticking to a version is what we have to do, debt tech, that we are paying now to control when we have to update;

microhobby: Ah also this crashed the WSL boot. So ... better to stick to a version and update only when neeeded

LeoBelmont: Ok, makes sense. I was mostly confused by this comment "make sure we are also updating to the latest version of xonsh" since it says latest version of xonsh. Maybe we should say "target version of xonsh"?

microhobby: yeah, I will change this comment

LeoBelmont (approved):

PR #465, opened 2026-06-24
Update the Slint templates to the latest 1.17.0 release

tronical:


microhobby:
Hey @tronical sorry abou the delay, I will accept this one I will already bump to the 1.17.1, if it's ok from your side

tronical:
Yes that would be greatly appreciated:)

